PR #16, opened 2015-02-11
Consumer deployment executor w/ proper consumer shutdown.

zdavep:
Fixes various issues found while investigating forklift.


connectors/activemq/src/main/java/forklift/connectors/ActiveMQConnector.java
```diff
@@ -99,7 +99,9 @@ public MessageConsumer getTopic(String name)
     public ForkliftMessage jmsToForklift(Message m) {
         try {
             final ForkliftMessage msg = new ForkliftMessage(m);
-            msg.setMsg(((ActiveMQTextMessage)m).getText());
+            if (m instanceof ActiveMQTextMessage) {
+                msg.setMsg(((ActiveMQTextMessage)m).getText());
+            }
```
dcshock: Probably should do some type of warning here if we can't handle the message type. Maybe a straight jms message passthrough the ForkliftMessage object.

PR #18, opened 2015-02-23
create docker container. add fasterxml dependency to build.sbt

paulCodes:
basic docker container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,33 @@
+# Forklift server Dockerfile
+# 2015-02-20
+
+# Pull base image
+FROM dockerfile/java:oracle-java8
+
+# Main developer
+MAINTAINER Matt Conroy <elduderino@mailinator.com>
+
+# Update system
+RUN apt-get update
+RUN apt-get upgrade -y
+
+# Install unzip
+RUN apt-get install unzip -y --force-yes --no-install-recommends
+
+# Add forklift server
+ADD ./server/target/universal/forklift-server-0.1.zip /tmp/forklift-server-0.1.zip
+WORKDIR /tmp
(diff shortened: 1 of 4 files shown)


core/build.sbt
```diff
@@ -20,6 +20,7 @@ libraryDependencies ++= Seq(
     "org.apache.geronimo.specs" % "geronimo-jms_1.1_spec" % "1.1.1",
     "org.reflections" % "reflections" % "0.9.9-RC1",
     "com.novocode" % "junit-interface" % "0.10" % "test",
+    "com.fasterxml.jackson.core" % "jackson-databind" % "2.5.1",
```
dcshock: Is there a reason this is being included with this PR?

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+# Update system
+RUN apt-get update
+RUN apt-get upgrade -y
+
+# Install unzip
+RUN apt-get install unzip -y --force-yes --no-install-recommends
+
+# Add forklift server
+ADD ./server/target/universal/forklift-server-0.1.zip /tmp/forklift-server-0.1.zip
```
dcshock: Can you add a doc file explaining how the 0.1.zip file is generated? Or better yet, the process to follow to generate the docker container?

core/build.sbt
zdavep: The forklift server won't start in a docker container without it.  Errors out with a ClassNotFound exception.

dcshock: Got it. Thanks for the info.

Dockerfile
paulCodes: added DOCKERCONTAINER.md

PR #19, opened 2015-03-25
Feature/more decorators for features

applitect:
Add some more of the decorators we'd like to support as well as some documentation about what we are trying to accomplish with this product. Still a work in progress but we'll get there.


applitect:
Oops merged to wrong branch

PR #27, opened 2015-04-28
Feature/deployment prop unit tests

applitect:


core/src/main/java/forklift/properties/PropertiesManager.java
```diff
@@ -65,6 +62,6 @@ public static Properties get(String name) {
     }
 
     private String getName(File f) {
-        return f.getName().split(".")[0];
+        return f.getName().split("\\.")[0];
```
dcshock: Nice Catch!

Issue #14, opened 2015-02-10
Producer Support


zdavep:
Feature Request - Add the ability to build producers in the same simple manner as consumers.

dcshock:
@paulCodes is working on this now.

PR #40, opened 2015-05-19
Feature/create producer

paulCodes:
* Create new Forklift Producer Interface
* Create Forklift Producer annotation so that you can inject a producer into a consumer
* Create ActiveMQProducer
* Create ActiveMQProducer junit Test
* Update ActiveMQ messaging test to use new ForkliftProducerI
* Update ActiveMQ Test connector to use new ForkliftProducerI
* Update ActiveMQConnector by adding the register method to inject ForkliftProducers into Consumers.

addresses issue #14


connectors/activemq/src/main/java/forklift/connectors/ActiveMQConnector.java
```diff
@@ -137,20 +139,45 @@ public ForkliftMessage jmsToForklift(Message m) {
...
         } catch (JMSException | ConnectorException e) {
+            System.out.println("getTopicProducer, throwing error");
             return null;
         }
     }
+
+    @Override
+    public void register(Object instance) {
+        try {
+            Arrays.stream(instance.getClass().getDeclaredFields())
```
dcshock: We should keep injection logic in the Consumer class. Thoughts?

paulCodes: Tried to keep it in the consumer, but could not find a way to inject the producer outside of the message loop. Kept running into the chicken/egg scenario.

paulCodes: Went ahead and refactored the injection out of the connector like we talked about, and kept it inside of the onMessage loop.

paulCodes:
Once this passes the checks it should ready to merge in.

connectors/activemq/src/main/java/forklift/producers/ActiveMQProducer.java
```diff
@@ -0,0 +1,223 @@
...
+            }
+            setMessageHeaders(msg, message.getHeaders());
+            setMessageProperties(msg, message.getProperties());
+            return msg;
+        } catch (Exception e) {
+            throw new ProducerException("Error creating JMS Message, Forklift message was null", e);
+        }
+    }
+
+    private void setMessageHeaders(Message msg, Map<Header, Object> headers) throws ProducerException {
```
applitect: This can be changed now with the ActiveMQHeaders settters.

PR #44, opened 2015-05-28
producer sends objects

paulCodes:
* add overload to producer so you can send an object and the producer will handle the marshaling into a forklift message.
* add some javadoc to the send methods


dcshock:
@paulCodes - can you look at the error?

paulCodes:
Ready to merge

PR #51, opened 2015-06-15
add condition to suppress json logging in dev

paulCodes:
no more impossible to read logs in dev huzzah!


dcshock:
@paulCodes does this work without any additional dependencies in the build.sbt file?

dcshock:
Another update trumps this one, and provides the same functionality.

PR #52, opened 2015-06-15
Feature: @On(ProcessStep) hooks

jgsofi:
Each consumer can now have custom callbacks for steps for notification and chaining.


core/src/main/java/forklift/consumer/MessageRunnable.java
```diff
@@ -42,51 +44,62 @@
...
-                        LifeCycleMonitors.call(ProcessStep.Invalid, this);
-                    } else {
-                        LifeCycleMonitors.call(ProcessStep.Processing, this);
-                        for (Method m : onMessage) {
-                            // Send the message to each handler.
-                            m.invoke(handler, new Object[] {});
-                        }
+                // Run the message if there are no errors.
+                if (errors.size() > 0) {
+                    runHooks(ProcessStep.Invalid);
```
dcshock: This won't output the errors to the log anymore.

jgsofi: That is a pretty serious regression, should be fixed now.

core/src/main/java/forklift/consumer/MessageRunnable.java
```diff
@@ -42,51 +44,67 @@
...
-                            // Send the message to each handler.
-                            m.invoke(handler, new Object[] {});
-                        }
+                // Run the message if there are no errors.
+                if (errors.size() > 0) {
+                    // { Invalid }
+                    getErrors().stream().forEach(e -> log.error(e));
+                    runHooks(ProcessStep.Invalid);
+                    // !!EXIT!! if invalid, do not continue
+                    return;
```
dcshock: This will cause a message to not be ack'd and the queue will lock up. We need to ensure that any "handled" message is ack'd.

jgsofi:
This is a rewrite of the core lifecycle handling logic to hopefully make the logic flow more obvious and easy to verify, by wrapping up error handling into a helper function and limiting control flow to if statements.

----

There were some failing tests after the rework.

The first test was ensuring that the message handling resulted in an Error when random nulls were passed in the constructor. While I could restore this behavior, it really cannot (or should not) happen outside this tes. (The only place MessageRunnable is constructed is in Consumer, which provides non-null arguments.)

The other two tests were verifying that errors thrown before Processing result in an Error state (after passing through Invalid) which is a behavior that I am fixing by turning Invalid into a finishing state.

jgsofi:
This PR has become a bit of a mess. Ill rebase it, fix it up, and retry.

PR #61, opened 2015-07-01
Allow non-JSON logging to STDOUT if DEV_LOGGING is set

nateww:
- Janino libraries are already provided, so this is a no-op for most environments


dcshock:
merge develop rather than master.

PR #62, opened 2015-07-01
Remove Scala dependencies from libraries since this is a pure-Java project

nateww:


dcshock:
@nateww can you change the merge to location to develop rather than master?

Issue #49, opened 2015-06-15
ProcessStep:Invalid Should Be a Final State


dcshock:
Jake pointed this out last week, and it needs to be fixed. Invalid messages still show as error which makes the life cycle inaccurate to the documentation.

PR #68, opened 2015-07-14
@On(ProcessStep) annotation hooks

jgsofi:
# Description

Adds an @On annotation that allows Consumers to write triggers for arbitrary lifecycle events, including documentation.

# Lifecycle ordering

The lifecycle has been reordered to match the documentation. This should close [this bug](https://github.com/dcshock/forklift/issues/49).

# Plugins

The pgp sbt plugin was added to a few packages where it is needed to build, but not included (presumably assuming that it is included in the global sbt settings of the user.

# Lifecycle handling rewrite

The last patch features a rewrite of the core lifecycle handling logic to hopefully make the logic flow more obvious and easy to verify, by wrapping up error handling into a helper function and limiting control flow to if statements.

The main invariants of the function:
- If there are any errors before Processing begins, the resulting state is Invalid.
- If there are any errors after Processing begins, the resulting state is Error.
- If there are no errors, the resulting state is Complete.
- Always log errors and close JMS.

There were some failing tests after the rework.

The first test was ensuring that the message handling resulted in an Error when random nulls were passed in the constructor. While I could restore this behavior, it really cannot (or should not) happen outside this test. (The only place MessageRunnable is constructed is in Consumer, which provides non-null arguments.)

The other two tests were verifying that errors thrown before Processing result in an Error state (after passing through Invalid) which is a behavior that I am fixing by turning Invalid into a finishing state.

# Old PR

There was a previous PR that was a tragic mess, and has been dropped in favor of this one, but for the sake of preserving the development history properly, [here it is](https://github.com/dcshock/forklift/pull/52).


core/src/main/java/forklift/consumer/MessageRunnable.java
```diff
@@ -41,52 +43,50 @@
...
+            if (errors.size() > 0) {
+                // { Invalid }
+                runHooks(ProcessStep.Invalid);
+                LifeCycleMonitors.call(ProcessStep.Invalid, this);
+            } else {
+                // { Processing }
+                runHooks(ProcessStep.Processing);
+                LifeCycleMonitors.call(ProcessStep.Processing, this);
+                for (Method m : onMessage) {
+                    runLoggingErrors(() -> m.invoke(handler));
```
dcshock: For some reason the try/catch of the Throwable has been removed. So in the case that the message handler throws an NPE the consumer thread will die, and then restart. Also the message would be lost because it was Ack'd without completely processing through the lifecycle.

dcshock: Never mind. I see the functional interface now. Not too shabby.

core/src/main/java/forklift/consumer/MessageRunnable.java
```diff
@@ -124,4 +116,29 @@ public Object getHandler() {
...
+    private interface DangerousSupplier<T> {
+        T get() throws Throwable;
+    }
+    private <T> T runLoggingErrors(DangerousSupplier<T> func) {
+        try {
+            return func.get();
+        } catch (Throwable e) {
+            if (e.getCause() == null) {
+                addError(e.getMessage());
+            }
```
dcshock: Let's snuggle or get rid of the braces.

PR #73, opened 2015-07-21
Enhanced @Headers, @Properties, @Config

jgsofi:
This is a fix for issue 65. For documentation see the docs.

Gist with example code https://gist.github.com/jgsofi/04e37b61982c41316bad


core/src/main/java/forklift/consumer/Consumer.java
```diff
@@ -292,20 +297,72 @@ public void inject(ForkliftMessage msg, final Object instance) {
...
+                                if (config == null) {
+                                    log.warn("Attempt to inject field failed because resource file {} was not found", annotation.value());
+                                    return;
+                                }
+                                String key = annotation.field();
+                                if (key.equals("")) {
+                                    key = f.getName();
+                                }
+                                Object value = config.get(key);
+                                if (value != null) {
```
dcshock: Is there a reason to check for a null value rather than just setting the field? Eliminates a little code for a bad wired dependency.

core/src/main/java/forklift/message/Header.java
```diff
@@ -161,14 +161,15 @@
     Type("JMSType", "type", "message-annotations.x-opt-jms-type", String.class);
 
 
-    private String jmsMessage;
-    private String stompMessage;
-    private String amqpMessage;
-    private Type headerType;
+    final private String jmsMessage;
+    final private String stompMessage;
+    final private String amqpMessage;
```
dcshock: normally the final qualifier would come after private based on some arbitrary style I'm trying to uphold. :\

dcshock:
Added a couple comments, and need @jgsofi to resolve conflicts before we can move this one forward.

core/src/main/java/forklift/consumer/Consumer.java
jgsofi: Yes. The reason to check for null is to support "default values". `if key exists, set value, otherwise, use the existing default value`.

core/src/main/java/forklift/message/Header.java
jgsofi: Yeah, my bad, ill fix it right away.

dcshock:
Great addition, thanks @jgsofi

PR #75, opened 2015-07-30
Add org.w3c to jre delegation list

ironroot:
We needed to add org.w3c to the list of classes that we limit to system class loader

### DIFF
--- a/core/src/main/java/forklift/classloader/ChildFirstClassLoader.java
+++ b/core/src/main/java/forklift/classloader/ChildFirstClassLoader.java
@@ -26,9 +26,8 @@ public class ChildFirstClassLoader extends URLClassLoader {
 
         // There are some classes that we can't go to the child first to resolve. In these instances we'll intercept the call
         // and delegate it to the system classloader. All of these relate to base jre libraries that are bundled in java.
-        List<String> things = Arrays.asList("java", "com.sun", "sun", "org.xml", "org.w3c");
-
-        if(things.stream().filter(s -> name.startsWith(s)).count() > 0){
+        List<String> libraries = Arrays.asList("java", "com.sun", "sun", "org.xml", "org.w3c");
+        if (libraries.stream().anyMatch(s -> name.startsWith(s))) {
             try {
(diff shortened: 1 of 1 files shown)


core/src/main/java/forklift/classloader/ChildFirstClassLoader.java
```diff
@@ -17,57 +18,59 @@ public ChildFirstClassLoader(URL[] classpath, ClassLoader parent) {
...
+    @Override
+    protected synchronized Class<?> loadClass(String name, boolean resolve)
+      throws ClassNotFoundException {
+        // First, check if the class has already been loaded
+        Class<?> c = findLoadedClass(name);
 
         // There are some classes that we can't go to the child first to resolve. In these instances we'll intercept the call
         // and delegate it to the system classloader. All of these relate to base jre libraries that are bundled in java.
-        if (name.startsWith("java") || name.startsWith("com.sun") || name.startsWith("sun") || name.startsWith("org.xml")) {
+        List<String> things = Arrays.asList("java", "com.sun", "sun", "org.xml", "org.w3c");
```
dcshock: Let's use a more descriptive name than things.

core/src/main/java/forklift/classloader/ChildFirstClassLoader.java
```diff
@@ -17,57 +18,59 @@ public ChildFirstClassLoader(URL[] classpath, ClassLoader parent) {
...
+
+        if(things.stream().filter(s -> name.startsWith(s)).count() > 0){
```
dcshock: Rather than filter the list, let's use .anyMatch with a boolean return. This will avoid the count test, and having to iterate through the whole list.

paulCodes: We should still use things as a var name...

ironroot: I will make the change to .anyMatch and I will let you and Paul fight it out for the var name :)

Issue #65, opened 2015-07-08
Property Field Injection


dcshock:
Allow properties to be injected at the individual value level too. 

```
@Property("name") 
private String name;
```

rather than 

```
@Properties
private Properties props;

props.get("name");
```

Just seems to be an easy feature that is very functional.

PR #76, opened 2015-08-03
Allow properties to be injected at the individual value level

paulCodes:
fixes issue #65


dcshock:
:) - I think this is fixed in @jgsofi 's PR too.

paulCodes:
I pulled it down and tested it out, was getting exceptions.

paulCodes:
addressed in PR jgsofi:feature/65-property-field-injection

PR #87, opened 2016-02-12
Feature/add forklift gui

ShaneIsrael:
It does stuff and things


plugins/gui/bin/www
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env node
+
+/**
+ * Module dependencies.
+ */
+
+var app = require('../app');
+var debug = require('debug')('Noise:server');
```
dcshock: Noise shouldn't be referenced.

plugins/gui/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM build.sofi.com:5000/sofi-node-base
```
dcshock: This can't be built outside of the SoFi docker repo. Can you build a public node base container?

dcshock:
For the most part the interface is headed where it should be. Let's correct the SoFi and noise dependencies before merging. Once those are cleaned up I'll get this merged down with some minor cleanups.

ShaneIsrael:
Alright sounds good.

PR #91, opened 2016-03-22
Changed default port to 3001

ShaneIsrael:
This is so that it can run alongside Noise in the docker compose environment.


dcshock:
This isn't necessary. Just set the PORT environment variable rather than modifying the default.

ShaneIsrael:
Well okay then...

PR #99, opened 2016-04-15
- Add the required system implementation. Next up tests to confirm th…

dcshock:
…is works as expected.


dcshock:
@applitect this was ready, but I never did finish the test cases. I'll dig into this and get it finalized to close the PR.

PR #101, opened 2016-04-18
Feature/fix infinite loading confusion

ShaneIsrael:
Removed console logs, default port is now 3001, stopped infinite loop loading symbol.


plugins/gui/bin/start
```diff
@@ -8,7 +8,7 @@ var child = new forever.Monitor('./bin/www', {
     options: [],
     env: {
         TZ: process.env.TZ || "America/New_York",
-        PORT: process.env.PORT || 3000
+        PORT: process.env.PORT || 3001
```
dcshock: Why are u changing the default node port? This should remain 3000.

PR #103, opened 2016-04-20
More cleanup

vultron81:


connectors/activemq/src/main/java/forklift/producers/ActiveMQProducer.java
```diff
@@ -22,7 +22,8 @@
 
 public class ActiveMQProducer implements ForkliftProducerI {
 
-    private static final ObjectMapper mapper = new ObjectMapper();
+    private static final ObjectMapper mapper = new ObjectMapper().registerModule(new JavaTimeModule())
+                                                                 .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);;
```
dcshock: Double semicolon.

core/src/main/java/forklift/consumer/MessageRunnable.java
```diff
@@ -144,8 +144,8 @@ public void run() {
                     LifeCycleMonitors.call(ProcessStep.Complete, this);
                 }
             }
-            // Always log all errors
-            getErrors().stream().forEach(e -> log.error(e));
+            // Always log all non-null errors
+            getErrors().stream().filter(e -> e != null).forEach(e -> log.error(e));
```
dcshock: I'm wondering if it would be useful at this point to wrap the stacktrace into an info message so that it can be reviewed. Thoughts?

plugins/retry/src/main/java/forklift/retry/RetryHandler.java
```diff
@@ -47,8 +49,9 @@ public RetryHandler(ForkliftConnectorI connector) {
     public RetryHandler(ForkliftConnectorI connector, File dir) {
         this.dir = dir;
         this.connector = connector;
-        this.mapper = new ObjectMapper();
-        this.mapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
+        this.mapper = new ObjectMapper().registerModule(new JavaTimeModule())
+                                        .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+                                        .configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
```
dcshock: I'm on the fence on this one. The idea behind retries is that a message will be re-executed at a later time exactly the same as the first time. If the bean can't be converted in full it's possible we lose some data that would affect message processing.

dcshock:
Just a couple comments @vultron81 - Thx.

plugins/retry/src/main/java/forklift/retry/RetryHandler.java
vultron81: This was a weird one.  Even though my POJO had the JSON Ignore annotation, it refused to honor it and would through the a warning and exception.   This was the only way I could fix it.

PR #106, opened 2016-06-06
Sends out daily summary at specified time

ShaneIsrael:
Please review over this. I sent you environment variables on hipchat.


plugins/gui/mail/mailer.js
```diff
@@ -0,0 +1,118 @@
...
+        }
+        logger.info("sending forklift replay status email now");
+        sendReplayStatusEmail(message);
+    }, function(err) {
+        logger.error(err.message);
+    });
+}
+
+var sendReplayStatusEmail = function (message) {
+    request('http://' + ip + ':8500/v1/catalog/semairvice/activemq-broker', function (error, response, body) {
```
dcshock: This URL seems bad... Does this actually work?

dcshock:
It doesn't appear that the Cron job is ever created in this PR?

ShaneIsrael:
Forgot to wire the cron job, its in now. The URL is the same one that I use in open sesame for getting the activemq details. it uses the CORE_IP var which is the ip address of the box its running on.

plugins/gui/mail/mailer.js
```diff
@@ -0,0 +1,118 @@
...
+var fs = require('fs');
+var elasticsearch = require('elasticsearch');
+var logger = require('../utils/logger');
+var request = require('request');
+var stompit = require('stompit');
+
+var m = {}
+
+var ip = process.env.CORE_IP || "localhost";
+var to = process.env.FG_MAILING_LIST;
```
dcshock: If the to/from isn't available just return and don't start the job.

ShaneIsrael:
This is ready, just waiting on Trevor to implement my provided JWT bash function for the daily cron jobs once thats done we should be able to deploy forklift-gui and have a JWT authenticated endpoint which will kick off replay summary emails to anyone in that list.

ShaneIsrael:
Trevor and I have got the JWT stuff in place, we got a Cron job going with a template built for open sesame now. We can use that same template to start this JWT authenticated job to send out emails whenever you're ready for a new release on forklift-gui.

PR #113, opened 2016-10-05
- Transport.

dcshock:


plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
```diff
@@ -1,42 +1,29 @@
...
     private static final Logger log = LoggerFactory.getLogger(ReplayES.class);
 
-    private final JestClient client;
+    private final TransportClient client;
 
     public ReplayESWriter(boolean ssl, String hostname) {
         this(ssl, hostname, 9200);
     }
 
     public ReplayESWriter(boolean ssl, String hostname, int port) {
```
ironroot: ssl is not used at all in this method

dcshock: Correct, it's left for backward compatibility. Later on we'll implement it.

plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
```diff
@@ -46,39 +33,28 @@ protected void poll(ReplayESWriterMsg t) {
...
+                .actionGet();
+
+            if (resp != null && resp.getHits() != null && resp.getHits().getHits() != null) {
+                for (SearchHit hit : resp.getHits().getHits()) {
+                    client.prepareDelete(hit.getIndex(), "log", t.getId()).execute().actionGet();
+                }
             }
-        } catch (IOException e) {
+        } catch (Exception e) {
+            log.error("", e);
```
dcshock: The exception thrown here is really clear... ES no like us.. 🗡

PR #117, opened 2016-11-15
Feature/allow class deployments

AFrieze:
I think this is needed?  I incremented version numbers and added a note about the new version in the documentation.


applitect:
Very out of date pull request and only logic change doesn't appear to be right.

PR #118, opened 2016-11-16
SOFI-14088 Make Forklift-Gui Great Again

ShaneIsrael:
Spent the evening (after work) re-writing the gui. Peters team is using it more and more often and the tool previously was an embarrassment to my ability. It now has the potential to be an insanely useful tool for forklift since adding additional features will be easier. Its much cleaner now and looks professional. Some of the code for the log message design was borrowed from my personal website https://coopfriendfinder.com if that matters at all. They both happen to have a similar layout and so killed two birds with one stone. 

- Fixed the copy/paste issue by re-writing everything. 
- Properly using jade/pug template engine now makes the gui super fast. 
- Because of the above I increased the max messages from 100 to 500. Could probably do well over 1000 if it came down to it.

screen shot 2016-11-15 at 9 48 50 pm

screen shot 2016-11-15 at 9 46 21 pm

screen shot 2016-11-15 at 9 47 41 pm

@dcshock Let me know if theres any issues with this. Would be great if we could get it out ASAP.


ShaneIsrael:
Touched a couple minor things up, should be all set. please ship it. 👍

PR #124, opened 2017-03-06
Request review and merge of Kafka Connector for forklift project

AFrieze:
Kafka connector with confluent schema-registry for forklift.  Adapts to JMS specification and should be able to replace existing activemq producers and consumer in the majority of scenarios.


kaiserleib:
Failing in CI, just FYI

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,141 @@
...
+        this.parsedJsonSchema = parser.parse(jsonSchema);
+    }
+
+    @Override
+    public String send(String message) throws ProducerException {
+        GenericRecord avroRecord = new GenericData.Record(parsedStringSchema);
+        avroRecord.put("forkliftMsg", message);
+        ProducerRecord record = new ProducerRecord<>(topic, null, avroRecord);
+        try {
+            kafkaProducer.send(record);
```
Kuroshii: We should probably have a callback on all of the `kafkaProducer.send` calls to at least _log_ errors returned from the broker.

connectors/kafka/build.sbt
```diff
@@ -0,0 +1,82 @@
+organization := "com.github.dcshock"
+
+name := "forklift-kafka"
+
+version := "0.1"
+
+// target and Xlint cause sbt dist to fail
+javacOptions ++= Seq("-source", "1.8")//, "-target", "1.8", "-Xlint")
```
Kuroshii: You can still have linting on just compile with `javacOptions in (Compile, compile)`.

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,167 @@
...
+    private KafkaController createController() {
+        Properties props = new Properties();
+        props.put("bootstrap.servers", kafkaHosts);
+        props.put("group.id", groupId);
+        props.put("enable.auto.commit", false);
+        props.put("consumer.timeout.ms", "-1");
+        props.put("key.deserializer", io.confluent.kafka.serializers.KafkaAvroDeserializer.class);
+        props.put("value.deserializer", io.confluent.kafka.serializers.KafkaAvroDeserializer.class);
+        props.put("schema.registry.url", schemaRegistries);
+        props.put("specific.avro.reader", false);
```
Kuroshii: I think for most of our cases moving forward we would prefer to read avro as a `SpecificRecord` for the stronger type checking. Accessing data "as a map" is usually too flexible.

Also, it seems strange to support writing specific records, but not reading specific records.

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,167 @@
...
+    private final String kafkaHosts;
+    private final String schemaRegistries;
+    private final String groupId;
+    private KafkaConsumer<?, ?> kafkaConsumer;
+    private KafkaProducer<?, ?> kafkaProducer;
+    private MessageStream messageStream = new MessageStream();
+
+    static ObjectMapper mapper = new ObjectMapper().registerModule(new JavaTimeModule())
+                                                   .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+    KafkaController controller;
```
Kuroshii: Is there a reason to have these two variables package-local? Just for testing?

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Override
+    public ForkliftProducerI getQueueProducer(String name) {
+        return this.getTopicProducer(name);
+    }
+
+    @Override
+    public ForkliftProducerI getTopicProducer(String name) {
+        return new KafkaForkliftProducer(name, this.kafkaProducer);
```
Kuroshii: What if the `kafkaProducer` is `null` when this is called (i.e. `start` has not been called yet, or is in the process of being called)? Maybe an `IllegalStateException` would be appropriate.

But that's seems a little weird because you can somehow get a consumer without starting the connector but you can't get a producer without starting the connector.

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Override
+    public ForkliftConsumerI getTopic(String name) throws ConnectorException {
+        synchronized (this) {
+            if (this.controller == null || !this.controller.isRunning()) {
+                this.controller = createController();
+                this.controller.start();
+            }
+        }
+        return new KafkaTopicConsumer(name, controller, messageStream);
```
Kuroshii: Design nit: I think it would make more sense to have the controller manage the details of creating a consumer. That way you can put the logic for setting up a consumer to use a controller in the controller rather than having the consumer manage state to keep track of whether the controller is actually completely set up for it. This also has the minor benefit of avoiding exposing the `MessageStream` object to the connector if it doesn't need to be exposed.

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,141 @@
...
+                avroRecord.put("forkliftJsonMsg", json);
+                record = new ProducerRecord<>(topic, null, avroRecord);
+            }catch(Exception e){
+                throw new ProducerException("Error creating Kafka Message", e);
+            }
+        }
+        try {
+            kafkaProducer.send(record);
+        } catch (SerializationException e) {
+            throw new ProducerException("Error creating Kafka Message", e);
```
Kuroshii: Copy-pasted exception? I don't think this message makes sense here

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,141 @@
...
+        return null;
+    }
+
+    @Override
+    public String send(ForkliftMessage message) throws ProducerException {
+        return this.send(message.getMsg());
+    }
+
+    @Override
+    public String send(Object message) throws ProducerException {
```
Kuroshii: Minor: this method could use a little style clean up

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,141 @@
...
+ */
+public class KafkaForkliftProducer implements ForkliftProducerI {
+
+    private final String topic;
+    private final KafkaProducer<?, ?> kafkaProducer;
+    static ObjectMapper mapper = new ObjectMapper().registerModule(new JavaTimeModule())
+                                                   .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+    String stringSchema = "{\"type\":\"record\"," +
+                          "\"name\":\"ForkliftStringMessage\"," +
+                          "\"fields\":[{\"name\":\"forkliftMsg\",\"type\":\"string\"}]}";
```
Kuroshii: Avro has a map type; can we use that instead of a string?

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,141 @@
...
+
+    String mapSchema = "{\"type\":\"record\"," +
+                       "\"name\":\"ForkliftMapMessage\"," +
+                       "\"fields\":[{\"name\":\"forkliftMapMsg\",\"type\":\"string\"}]}";
```
Kuroshii: Avro has a map type; can we use that instead of a string?

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,167 @@
...
+            final ForkliftMessage msg = new ForkliftMessage(m);
+            if (m instanceof KafkaMessage) {
+                KafkaMessage kafkaMessage = (KafkaMessage)m;
+                msg.setProperties(kafkaMessage.getProperties());
+                ConsumerRecord<?, ?> record = kafkaMessage.getConsumerRecord();
+                if (record.value() instanceof GenericRecord) {
+                    GenericRecord genericRecord = (GenericRecord)record.value();
+                    Object value = genericRecord.get("forkliftMapMsg");
+                    value = value == null ? genericRecord.get("forkliftMsg") : value;
+                    value = value == null ? genericRecord.get("forkliftJsonMsg") : value;
```
Kuroshii: Shouldn't we deserialize these values from our message so that our message's value appears to be what was sent? I don't see anything deserializing the string that was constructed in the producer for maps, for example.

Also, with removing the JMS stuff, this is going to need to find a new home...

connectors/kafka/src/main/java/forklift/connectors/KafkaController.java
```diff
@@ -0,0 +1,225 @@
...
+                if (updatedAssignment) {
+                    this.acknowlegmentHandler.addPartitions(kafkaConsumer.assignment());
+                }
+                if (records.count() > 0) {
+                    log.debug("Adding: " + records.count() + " to record stream");
+                    messageStream.addRecords(consumerRecordsToKafkaMessages(records));
+                }
+                Map<TopicPartition, OffsetAndMetadata> offsetData = this.acknowlegmentHandler.flushAcknowledged();
+                if (offsetData.size() > 0) {
+                    String offsetDescription =
```
Kuroshii: I would rather not do this string construction all the time if it only matters for debug logs.

It would be cool if we could just throw `lazy` before this and call it good, but alas.

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
AFrieze: The string is parsed to the appropriate type based on the annotated Message object.  Look at the inject method in Consumer.java

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
AFrieze: I didn't think it was supported by our serializer

connectors/kafka/src/main/java/forklift/connectors/KafkaController.java
AFrieze: +1

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
Kuroshii: What do you mean? It's definitely in the avro spec, at least...

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
Kuroshii: I see, thanks.

connectors/kafka/build.sbt
```diff
@@ -0,0 +1,82 @@
...
+    sys.error("Java 8 is required for this project.")
+}
+
+libraryDependencies ++= Seq(
+  "com.github.dcshock" % "forklift" % "0.19" ,
+  "com.fasterxml.jackson.core" % "jackson-databind" % "2.7.3",
+  "com.fasterxml.jackson.datatype" % "jackson-datatype-jsr310" % "2.7.3",
+  "org.apache.kafka" % "kafka-clients" % "0.10.1.1-cp1",
+  "io.confluent" % "kafka-avro-serializer" % "3.1.1"
+
```
sagersmith8: Extra newline

connectors/kafka/build.sbt
```diff
@@ -0,0 +1,82 @@
...
+)
+
+lazy val testDependencies = Seq(
+  "commons-io" % "commons-io" % "2.4" ,
+  "com.novocode" % "junit-interface" % "0.11",
+  "org.mockito"       % "mockito-core"            % "1.9.5"
```
sagersmith8: Weird extra spaces should be:

    "org.mockito" % "mockito-core" % "1.9.5"

connectors/kafka/build.sbt
```diff
@@ -0,0 +1,82 @@
...
+crossPaths := false
+
+publishMavenStyle := true
+
+publishTo := {
+  val nexus = "https://oss.sonatype.org/"
+  if (isSnapshot.value)
+    Some("snapshots" at nexus + "content/repositories/snapshots")
+  else
+    Some("releases"  at nexus + "service/local/staging/deploy/maven2")
```
sagersmith8: Extra space, should be:

    Some("releases" at nexus + "service/local/staging/deploy/maven2")

connectors/kafka/build.sbt
```diff
@@ -0,0 +1,82 @@
...
+autoScalaLibrary := false
+
+// Remove the scala version from the generated/published artifact
+crossPaths := false
+
+publishMavenStyle := true
+
+publishTo := {
+  val nexus = "https://oss.sonatype.org/"
+  if (isSnapshot.value)
```
sagersmith8: Best not to use single line if

connectors/kafka/build.sbt
```diff
@@ -0,0 +1,82 @@
...
+    Some("snapshots" at nexus + "content/repositories/snapshots")
+  else
```
sagersmith8: Best not to use single line else

connectors/kafka/src/main/java/forklift/connectors/AcknowledgedRecordHandler.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Maintains a batch of acknowledged records and provides management for adding and removing {@link org.apache.kafka.common.TopicPartition partitions}.
+ * Acknowledged records are those that have started processing but have not yet been committed to the Kafka Broker.  This class is threadsafe.
+ */
+public class AcknowledgedRecordHandler {
+    private static final Logger log = LoggerFactory.getLogger(AcknowledgedRecordHandler.class);
```
sagersmith8: Should be named `LOGGER` per java convention on `static final` varaibles

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,167 @@
...
+public class KafkaConnector implements ForkliftConnectorI {
+    private static final Logger log = LoggerFactory.getLogger(KafkaConnector.class);
+    private final String kafkaHosts;
+    private final String schemaRegistries;
+    private final String groupId;
+    private KafkaConsumer<?, ?> kafkaConsumer;
+    private KafkaProducer<?, ?> kafkaProducer;
+    private MessageStream messageStream = new MessageStream();
+
+    static ObjectMapper mapper = new ObjectMapper().registerModule(new JavaTimeModule())
```
sagersmith8: This could be final

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
sagersmith8: Yea I could prefer to see these be `protected` because of general confusion around the `default` modifier if the access modifier is just for testing

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,167 @@
...
+
+    public KafkaConnector(String kafkaHosts, String schemaRegistries, String groupId) {
+        this.kafkaHosts = kafkaHosts;
+        this.schemaRegistries = schemaRegistries;
+        this.groupId = groupId;
+    }
+
+    @Override
+    public void start() throws ConnectorException {
+
```
sagersmith8: Extra newline

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,167 @@
...
+        producerProperties.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG,
+                               io.confluent.kafka.serializers.KafkaAvroSerializer.class);
+        //schema.registry.url is a comma separated list of urls
+        producerProperties.put("schema.registry.url", schemaRegistries);
+        this.kafkaProducer = new KafkaProducer(producerProperties);
+        this.controller = createController();
+        this.controller.start();
+    }
+
+    private KafkaController createController() {
```
sagersmith8: Missing javadocs

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,167 @@
...
+        Properties props = new Properties();
+        props.put("bootstrap.servers", kafkaHosts);
+        props.put("group.id", groupId);
+        props.put("enable.auto.commit", false);
+        props.put("consumer.timeout.ms", "-1");
+        props.put("key.deserializer", io.confluent.kafka.serializers.KafkaAvroDeserializer.class);
+        props.put("value.deserializer", io.confluent.kafka.serializers.KafkaAvroDeserializer.class);
+        props.put("schema.registry.url", schemaRegistries);
+        props.put("specific.avro.reader", false);
+        //props.put("auto.offset.reset", "earliest");
```
sagersmith8: If you don't need this you should remove

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Override
+    public void stop() throws ConnectorException {
+        try {
+            this.controller.stop(2000, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+            log.error("KafkaConnector interrupted while stopping");
+        }
+        this.kafkaProducer.close();
```
sagersmith8: Probably wrap this in a `finally` is my guess`

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,167 @@
...
+        return this.getTopicProducer(name);
+    }
+
+    @Override
+    public ForkliftProducerI getTopicProducer(String name) {
+        return new KafkaForkliftProducer(name, this.kafkaProducer);
+    }
+
+    @Override
+    public ForkliftMessage jmsToForklift(Message m) {
```
sagersmith8: No single character variables please, it makes debugging hard

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,167 @@
...
+                        value = jsonValue != null && jsonValue.startsWith("{") ? jsonValue : value;
+                    }
+                    if (value == null) {
+                        msg.setFlagged(true);
+                        msg.setWarning("Unable to parse message for topic: " + record.topic() + " with value: " + record.value());
+                    } else {
+                        msg.setMsg(value.toString());
+                    }
+                } else {
+                    ObjectMapper mapper = new ObjectMapper();
```
sagersmith8: Why don't you use the one that you declared at the top?

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
sagersmith8: Doesn't this line just overwrite the last line? This seems like you are trying to do a weird cascading check, but it won't actually tell you what went wrong if that is the case. Why don't you just use if checks?

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,167 @@
...
+                    try {
+                        //inefficient as we map from object to json, then back to the object.  This approach works
+                        //without any changes to the forklift core libraries however.
+                        msg.setMsg(mapper.writeValueAsString(record.value()));
```
sagersmith8: again why don't you use the ObjectMapper declared at the top>

AFrieze:
All -  Thank you for the feedback.  We have decided to keep the kafka functionality in a project specific to Kafka for now to allow us to make more changes to the core and move away from the JMS dependencies.  We may have a larger PR coming in for that in the future if that's the way forklift wants to go.  All the feedback is very much appreciated.

PR #125, opened 2017-03-30
Wip/forklift kafka

AFrieze:
Add support for kafka


connectors/activemq/src/main/java/forklift/consumer/ActiveMQMessageConsumer.java
```diff
@@ -16,15 +27,60 @@ public ActiveMQMessageConsumer(MessageConsumer consumer, Session s) {
 		this.s = s;
 	}
 
-	public Message receive(long timeout) throws JMSException {
-		return consumer.receive(timeout);
+	public ForkliftMessage receive(long timeout) throws ConnectorException {
```
Kuroshii: Nit: tabs

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,135 @@
...
+                               io.confluent.kafka.serializers.KafkaAvroSerializer.class);
+        //schema.registry.url is a comma separated list of urls
+        producerProperties.put("schema.registry.url", schemaRegistries);
+        kafkaProducer = new KafkaProducer(producerProperties);
+        return kafkaProducer;
+    }
+
+    private KafkaController createController() {
+        Properties props = new Properties();
+        props.put("bootstrap.servers", kafkaHosts);
```
Kuroshii: Seems weird to use string property names here and the constants for the producer.

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,135 @@
...
+        props.put("group.id", groupId);
+        props.put("enable.auto.commit", false);
+        props.put("key.deserializer", io.confluent.kafka.serializers.KafkaAvroDeserializer.class);
+        props.put("value.deserializer", io.confluent.kafka.serializers.KafkaAvroDeserializer.class);
+        props.put("schema.registry.url", schemaRegistries);
+        props.put("specific.avro.reader", false);
+        props.put("auto.offset.reset", "earliest");
+        props.put("max.poll.records", "200");
```
Kuroshii: `200` could probably just be an `int` constant.

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,135 @@
...
+        KafkaConsumer<?, ?> kafkaConsumer;
+        kafkaConsumer = new KafkaConsumer(props);
```
Kuroshii: Nit: doesn't need it's own line

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,135 @@
...
+
+        Properties producerProperties = new Properties();
+        producerProperties.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaHosts);
+        producerProperties.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG,
+                               io.confluent.kafka.serializers.KafkaAvroSerializer.class);
+        producerProperties.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG,
+                               io.confluent.kafka.serializers.KafkaAvroSerializer.class);
+        //schema.registry.url is a comma separated list of urls
+        producerProperties.put("schema.registry.url", schemaRegistries);
+        kafkaProducer = new KafkaProducer(producerProperties);
```
Kuroshii: Nit: the `kafkaProducer` field is set, returned, then set again from the same value.

connectors/kafka/build.sbt
```diff
@@ -0,0 +1,103 @@
...
+libraryDependencies ++= Seq(
+  "com.github.dcshock" % "forklift" % "1.0" ,
+  "com.fasterxml.jackson.core" % "jackson-databind" % "2.7.3",
+  "com.fasterxml.jackson.datatype" % "jackson-datatype-jsr310" % "2.7.3",
+  "org.apache.kafka" % "kafka-clients" % "0.10.1.1-cp1" exclude("org.slf4j","slf4j-log4j12"),
+  "io.confluent" % "kafka-avro-serializer" % "3.1.1" exclude("org.slf4j","slf4j-log4j12"),
+  "io.confluent" % "kafka-schema-registry" % "3.1.1" exclude("org.slf4j","slf4j-log4j12"),
+  "org.apache.avro" % "avro" % "1.8.1",
+  "org.apache.zookeeper" % "zookeeper" % "3.4.9" exclude("org.slf4j","slf4j-log4j12")
+
```
Kuroshii: Nit: blank line

connectors/kafka/build.sbt
```diff
@@ -0,0 +1,103 @@
...
+)
+
+
+// avro settings
+(javaSource in avroConfig) := baseDirectory(_/"target/generated-sources/").value
+(sourceDirectory in avroConfig) := baseDirectory(_/"src/schemas").value
+
+(compile in Compile) := ((compile in Compile) dependsOn (clean in Compile)).value
+
+libraryDependencies ++= testDependencies.map(_ % "test")
```
Kuroshii: Nit: strange to not put this immediately after `testDependencies`

connectors/kafka/src/schemas/UserRegistered.avsc
```diff
@@ -0,0 +1,31 @@
+{
+  "type" : "record",
+  "name" : "UserRegistered",
+  "namespace" : "com.sofi.avro.schemas",
```
Kuroshii: `com.sofi`

connectors/kafka/src/test/java/forklift/integration/AvroMessageTests.java
```diff
@@ -0,0 +1,87 @@
+package forklift.integration;
+
+import com.sofi.avro.schemas.StateCode;
+import com.sofi.avro.schemas.UserRegistered;
```
Kuroshii: Yeah, probably can just remove these...

connectors/kafka/src/schemas/Test.avsc
```diff
@@ -0,0 +1,10 @@
+{
```
Kuroshii: The avro-generated code is still appearing in the packaged jar; maybe you can get rid of it by making sbt `clean` before `publishLocal` and then making `avro`depend on `test:compile` instead of `compile`.

connectors/kafka/src/test/java/forklift/connectors/KafkaControllerTests.java
```diff
@@ -0,0 +1,143 @@
...
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import java.util.Collection;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Created by afrieze on 3/3/17.
+ */
```
Kuroshii: Can we just not have these b/c we have version control?

server/build.sbt
```diff
@@ -15,18 +15,17 @@ initialize := {
...
-  "com.github.dcshock" % "forklift-stats"     % "0.1",
+  "com.github.dcshock" % "forklift"           % "1.0",
+  "com.github.dcshock" % "forklift-activemq"  % "1.0",
+  "com.github.dcshock" % "forklift-replay"    % "1.0",
+  "com.github.dcshock" % "forklift-retry"     % "1.0",
+  "com.github.dcshock" % "forklift-stats"     % "1.0",
   "com.github.dcshock" % "consul-rest-client" % "0.10",
+  "org.apache.activemq" % "activemq-broker" % "5.14.0",
   "io.searchbox" % "jest" % "2.0.0",
   "org.apache.geronimo.specs" % "geronimo-jms_1.1_spec" % "1.1.1",
```
Kuroshii: I don't think the JMS spec is actually used anywhere in the server

connectors/kafka/src/test/java/forklift/connectors/KafkaControllerTests.java
AFrieze: agreed, I need to configure intellij to stop adding those.

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,133 @@
...
+import org.apache.kafka.clients.producer.ProducerConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Manages both consuming and producing events on the kafka message broker.
+ */
+public class KafkaConnector implements ForkliftConnectorI {
```
Kuroshii: Thought: it's maybe a bit misleading to call this a general "Kafka connector" when it requires a schema registry.

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,133 @@
...
+    private KafkaProducer createKafkaProducer() {
+
+        Properties producerProperties = new Properties();
+        producerProperties.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaHosts);
+        producerProperties.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG,
+                               io.confluent.kafka.serializers.KafkaAvroSerializer.class);
+        producerProperties.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG,
+                               io.confluent.kafka.serializers.KafkaAvroSerializer.class);
+        //schema.registry.url is a comma separated list of urls
+        producerProperties.put("schema.registry.url", schemaRegistries);
```
Kuroshii: I believe there is a constant for this one too.

AFrieze: It would most likely be in the confluent serializer package.  Do you know what it is? I haven't tried to hunt it down yet

Kuroshii: `AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG` from `io.confluent.kafka.serializers.AbstractKafkaAvroSerDeConfig`

AFrieze: (thumbsup)

connectors/kafka/src/schemas/Test.avsc
Kuroshii: Weird that the name of the schema file doesn't match the name of the schema.

connectors/kafka/src/schemas/UserRegistered.avsc
```diff
@@ -0,0 +1,31 @@
+{
```
Kuroshii: I'd actually prefer to just get rid of or modify this, because it seems weird to have something with such a direct purpose being used for a test.

plugins/retry/src/main/java/forklift/retry/RetryHandler.java
```diff
@@ -37,26 +35,26 @@
...
         this.dir = dir;
-        this.connector = connector;
+        this.forklift = forklift;
         this.mapper = new ObjectMapper().registerModule(new JavaTimeModule())
                                         .configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
         this.executor = Executors.newScheduledThreadPool(1);
 
         // Cleanup after a retry is completed.
-        cleanup = new Callback<RetryMessage>() {
+        cleanup = new Consumer<RetryMessage>() {
```
Kuroshii: Nit: If we're using java 8 functions, we might as well use lambdas

plugins/retry/src/main/java/forklift/retry/RetryES.java
```diff
@@ -105,9 +103,9 @@ public RetryES(ForkliftConnectorI connector, boolean ssl, String hostname, int p
         client = factory.getObject();
 
         // Cleanup after a retry is completed.
-        cleanup = new Callback<RetryMessage>() {
+        cleanup = new Consumer<RetryMessage>() {
```
Kuroshii: Nit: lambda thing again

plugins/replay/src/main/java/forklift/replay/ReplayES.java
```diff
@@ -77,8 +76,9 @@ public void run() {
         });
 
         this.producer = connector.getQueueProducer(ReplayConsumer.class.getAnnotation(Queue.class).value());
-
-        this.consumer = new Consumer(ReplayConsumer.class, connector,
+        final Forklift forklift = new Forklift();
```
Kuroshii: Why not reuse the `Forklift` instance from when this plugin was created like `RetryES` does?

plugins/notify/build.sbt
```diff
@@ -2,7 +2,7 @@ organization := "com.github.dcshock"
 
 name := "forklift-notify"
```
Kuroshii: This plugin doesn't do anything and nobody uses it, right? Why bother keeping it around?

connectors/activemq/src/main/java/forklift/producers/ActiveMQProducer.java
```diff
@@ -185,51 +204,48 @@ private Message prepAndValidate(ForkliftMessage message,
...
-                try {
-                    msg = session.createTextMessage(message.getMsg());
-                } catch (Exception e) {
-                    throw new ProducerException("Error creating new jms TextMessage", e);
-                }
-            }
+            final Message msg = session.createTextMessage(message.getMsg());
+
+            // Set the jms correlation id.
+            if (message.getId() != null && !message.getId().trim().equals(""))
```
Kuroshii: Opinion: `isEmpty()` is better than `equals("")`

connectors/activemq/src/main/java/forklift/producers/ActiveMQProducer.java
```diff
@@ -185,51 +204,48 @@ private Message prepAndValidate(ForkliftMessage message,
...
+                msg.setJMSCorrelationID(message.getId());
+
             setMessageHeaders(msg, message.getHeaders());
             setMessageProperties(msg, message.getProperties());
             return msg;
         } catch (Exception e) {
-            throw new ProducerException("Error creating JMS Message, Forklift message was null", e);
+            throw new ProducerException("Error creating JMS Message, or Forklift message was null", e);
```
Kuroshii: We might as well leave off the "Forklift message was null" bit. That should be fairly obvious when you get a NPE. If it really matters, I think it's better to check whether the message is null and throw just for that.

connectors/activemq/src/main/java/forklift/producers/ActiveMQProducer.java
```diff
@@ -185,51 +204,48 @@ private Message prepAndValidate(ForkliftMessage message,
...
     private void setMessageHeaders(Message msg, Map<Header, Object> headers) throws ProducerException {
         if (msg != null && headers != null) {
             headers.entrySet().stream().filter(entry -> entry.getValue() != null)
-                              .forEach(entry -> {
-                                if(ActiveMQHeaders.getFunctions().get(entry.getKey()).get((ActiveMQMessage)msg) == null) {
-                                   ActiveMQHeaders.getFunctions().get(entry.getKey()).set((ActiveMQMessage)msg,entry.getValue());
-                                }
-                            });
+                .forEach(entry -> {
+                if(ActiveMQHeaders.getFunctions().get(entry.getKey()).get((ActiveMQMessage)msg) == null) {
```
Kuroshii: Nit: indentation

connectors/activemq/src/main/java/forklift/producers/ActiveMQProducer.java
```diff
@@ -185,51 +204,48 @@ private Message prepAndValidate(ForkliftMessage message,
...
-                                                }
-                                            } catch (Exception e) {
-                                                //Catch it but just move on.
-                                            }
-                                           });
+            final Map<String, String> props = new HashMap<>();
+            if (properties != null)
+                props.putAll(properties);
+            if (this.properties != null)
+                props.putAll(this.properties);
```
Kuroshii: Weird to have producer properties override passed properties

connectors/activemq/src/main/java/forklift/producers/ActiveMQProducer.java
```diff
@@ -185,51 +204,48 @@ private Message prepAndValidate(ForkliftMessage message,
...
 
     private void setMessageHeaders(Message msg, Map<Header, Object> headers) throws ProducerException {
         if (msg != null && headers != null) {
             headers.entrySet().stream().filter(entry -> entry.getValue() != null)
-                              .forEach(entry -> {
-                                if(ActiveMQHeaders.getFunctions().get(entry.getKey()).get((ActiveMQMessage)msg) == null) {
-                                   ActiveMQHeaders.getFunctions().get(entry.getKey()).set((ActiveMQMessage)msg,entry.getValue());
-                                }
-                            });
+                .forEach(entry -> {
```
Kuroshii: Do we want to also factor in stored properties here, like in `setMessageProperties`?

connectors/activemq/src/test/java/forklift/activemq/test/TestServiceManager.java
```diff
@@ -29,7 +31,9 @@ public static void start() {
             try {
                 activemq = new BrokerService();
                 activemq.addConnector("tcp://127.0.0.1:61618");
+                activemq.setDataDirectoryFile(Files.createTempDir());
                 activemq.start();
+                System.out.println("HEY" + activemq.getDataDirectoryFile());
```
Kuroshii: Probably don't still need this test log

connectors/kafka/README.md
```diff
@@ -0,0 +1,21 @@
...
+### Queues and Topics ###
+Kafka only provides one form of topic which is analogous to ActiveMQ's Virtual Topic.  Only one member from
+each subscribed group will receive a message.  Requests for a queue or topic will return the 
+same producer type.
+
+### At Most Once Delivery ###
+The kafka connector does not guarantee at most once delivery although it makes a best effort. 
+In order to adapt to the JMS spec, messages are acknowledged and added to a pending commit batch, 
+which are committed to kafka every poll cycle.  It is therefore possible for the server to crash 
+before the acknowledgment batch has been committed, and after the message has processed.
```
Kuroshii: Isn't this at-least-once processing then?

connectors/kafka/README.md
```diff
@@ -0,0 +1,21 @@
...
+## JMS ##
+Forklift was developed against the JMS spec.  Kafka is not a JMS implementation so some adaptation
+has occurred in order to allow Kafka to function with Forklift.
+
+### Queues and Topics ###
+Kafka only provides one form of topic which is analogous to ActiveMQ's Virtual Topic.  Only one member from
+each subscribed group will receive a message.  Requests for a queue or topic will return the 
+same producer type.
+
+### At Most Once Delivery ###
```
Kuroshii: Minor: this section is a bit awkward and could use a little work.

This isn't that bad, but hard to understand documentation is sometimes worse than no documentation.

connectors/kafka/src/schemas/Test.avsc
Kuroshii: Maybe an better solution is to generate these avro sources into the `test` folder and add the whole generated package to the connector level `.gitignore`.

For instance:
- Move the schemas into `src/test/resources/schemas` (seems more correct)
- Modify `build.sbt`:
```
(javaSource in avroConfig) := baseDirectory(_/"src/test/java").value
(sourceDirectory in avroConfig) := baseDirectory(_/"src/test/resources/schemas").value
```
- Add `/src/test/java/com/github/dcshock/avro/schemas` (maybe there's a better package name? `forklift.schemas`?) to the kafka-connector `.gitignore`.

Only real downside to this is if we want to use avro java objects in the main connector, it becomes difficult (although I'm sure it's solvable).

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,133 @@
...
+        Properties props = new Properties();
+        props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaHosts);
+        props.put(ConsumerConfig.GROUP_ID_CONFIG, groupId);
+        props.put(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, false);
+        props.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, io.confluent.kafka.serializers.KafkaAvroDeserializer.class);
+        props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, io.confluent.kafka.serializers.KafkaAvroDeserializer.class);
+        props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+        props.put(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, 200);
+        props.put("schema.registry.url", schemaRegistries);
+        props.put("specific.avro.reader", false);
```
Kuroshii: The property name for this is `KafkaAvroDeserializerConfig.SPECIFIC_AVRO_READER_CONFIG`.


Default value is `false`, but either way is okay. :man_shrugging:

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,133 @@
...
+        KafkaConsumer<?, ?> kafkaConsumer = new KafkaConsumer(props);
+        return new KafkaController(kafkaConsumer, messageStream);
+    }
+
+    @Override
+    public synchronized void stop() throws ConnectorException {
+        try {
+            if (controller != null) {
+                this.controller.stop(2000, TimeUnit.MILLISECONDS);
```
Kuroshii: Minor nit: there are a lot of redundant `this` references lying around.

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,133 @@
...
+import org.slf4j.LoggerFactory;
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Manages both consuming and producing events on the kafka message broker.
+ */
+public class KafkaConnector implements ForkliftConnectorI {
+    private static final Logger log = LoggerFactory.getLogger(KafkaConnector.class);
+    private final String kafkaHosts;
```
Kuroshii: Minor nit: These fields could be separated a little better (i.e separate values from state)

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -0,0 +1,133 @@
...
+    private final String schemaRegistries;
+    private final String groupId;
+    private KafkaProducer<?, ?> kafkaProducer;
+    private MessageStream messageStream = new MessageStream();
```
Kuroshii: Is there any reason for the connector to hold a reference to a `MessageStream`?

You should be able to either get rid of it or make it `final`?

connectors/kafka/src/main/java/forklift/message/MessageStream.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A stream of available messages which can be retrieved by topic.
+ */
+public class MessageStream implements ReadableMessageStream {
+    private static final Logger log = LoggerFactory.getLogger(MessageStream.class);
+    Map<String, BlockingQueue<KafkaMessage>> topicQueue = new ConcurrentHashMap<>();
```
Kuroshii: Given its usage, this could probably be made `private` and `final`.

connectors/kafka/src/main/java/forklift/message/MessageStream.java
```diff
@@ -0,0 +1,70 @@
...
+     * <strong>Note:</strong> All passed in records must belong to a topic added through a call to {@link #addTopic(String)}.
+     *
+     * @param records the records to add
+     * @throws IllegalStateException if the capacity of a stream has been exceeded.
+     */
+    public void addRecords(Map<TopicPartition, List<KafkaMessage>> records) {
+        log.debug("Adding records to stream");
+        for (Map.Entry<TopicPartition, List<KafkaMessage>> entry : records.entrySet()) {
+            for (KafkaMessage message : entry.getValue()) {
+                BlockingQueue<KafkaMessage> queue = topicQueue.get(entry.getKey().topic());
```
Kuroshii: Minor optimization: grabbing the blocking queue for a given topic can be done in the higher loop and then the `queue != null` check can be done earlier as well.

connectors/kafka/src/main/java/forklift/message/MessageStream.java
```diff
@@ -0,0 +1,70 @@
...
+                if (queue != null) {
+                    queue.add(message);
+                }
```
Kuroshii: Maybe we should do something in these cases when `queue` is `null`? At least log an error or throw an runtime exception? I don't like the idea of silently dropping messages if there is a bug.

connectors/kafka/src/main/java/forklift/message/MessageStream.java
```diff
@@ -0,0 +1,70 @@
...
+        this.topicQueue.remove(topic);
+    }
+
+    @Override
+    public KafkaMessage nextRecord(String topic, long timeout) throws InterruptedException {
+        KafkaMessage message = null;
+        BlockingQueue<KafkaMessage> queue = topicQueue.get(topic);
+        if (queue != null) {
+            message = queue.poll(timeout, TimeUnit.MILLISECONDS);
+        }
```
Kuroshii: Same kind of thing here, but less serious.

It might be worth it to just throw an `IllegalArgumentException` here if the queue is `null` (didn't break anything unexpected for me, just the one `removeTopic` test).

connectors/kafka/src/main/java/forklift/message/KafkaMessage.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    @Override
+    public String getId() {
+        return consumerRecord.topic() + "-" + consumerRecord.partition() + "-" + consumerRecord.offset();
+    }
+
+    /**
+     * <strong>WARNING:</strong> Called from constructor
```
Kuroshii: Why is this warning here? It's clear to me that it's called from the constructor, but is there some other implication  to that?

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * Maintains a batch of acknowledged records and provides management for adding and removing {@link org.apache.kafka.common.TopicPartition partitions}.
+ * Acknowledged records are those that have started processing but have not yet been committed to the Kafka Broker.  This class is threadsafe.
+ */
+public class AcknowledgedRecordHandler {
```
Kuroshii: It's a little weird to have so much documentation on an internal class. 😺

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+        this.assignment.addAll(addedPartitions);
+    }
+
+    /**
+     * Remove partitions from management.  Any existing offsets for the removed partitions are returned.  Note that the offset is the highest
+     * acknowledged message's offset + 1 per kafka's specification of how to commit offsets.  Note that this
+     * is a blocking method as any threads which are currently acknowledging records are allowed to complete and any
+     * incoming threads are paused.
+     *
+     * @param removedPartitions the partitions to remove
```
Kuroshii: `removedPartitions` is a bad name -- they aren't yet removed

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+        } finally {
+            lock.writeLock().unlock();
+        }
+    }
+
+    /**
+     * Adds additional partitions to be managed.  Only added partitions can be
+     * {@link #acknowledgeRecord(org.apache.kafka.clients.consumer.ConsumerRecord) acknowledged}
+     *
+     * @param addedPartitions the partitions to add
```
Kuroshii: `addedPartitions` also is bad name; maybe just `partitions` is clearer, since the `add` bit is implied by the method name?

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * Returns the highest offsets of any acknowledged records.
+     * This is a blocking method as a short delay may occur while any threads which are currently acknowledging records are allowed to
+     * complete and any incoming threads are paused.
+     *
+     * @return a Map of the highest offset data for any acknowledged records
+     * @throws InterruptedException if interrupted
+     */
+    public Map<TopicPartition, OffsetAndMetadata> getAcknowledged() throws InterruptedException {
+        lock.writeLock().lock();
```
Kuroshii: Isn't this in essence just a reading method? Why do you need to stop writers for this?

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+        try {
+            Map<TopicPartition, OffsetAndMetadata> currentOffsets = new HashMap<>(pendingOffsets.size());
+            currentOffsets.putAll(pendingOffsets);
```
Kuroshii: Can you just use the copy constructor of `HashMap` e.g.
`return new HashMap<>(pendingOffsets);`

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+ * Acknowledged records are those that have started processing but have not yet been committed to the Kafka Broker.  This class is threadsafe.
+ */
+public class AcknowledgedRecordHandler {
+    private static final Logger log = LoggerFactory.getLogger(AcknowledgedRecordHandler.class);
+    private ConcurrentHashMap<TopicPartition, OffsetAndMetadata> pendingOffsets = new ConcurrentHashMap<>();
+    private Set<TopicPartition> assignment = ConcurrentHashMap.newKeySet();
+    private ReadWriteLock lock = new ReentrantReadWriteLock();
+
+    /**
+     * Acknowledges that a record has been received before processing begins.  True is returned if processing should occur else false.
```
Kuroshii: The "true is returned ..." bit is redundant here.

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+        lock.readLock().lock();
+        try {
+            boolean acknowledged;
+            TopicPartition topicPartition = new TopicPartition(record.topic(), record.partition());
+            if (!assignment.contains(topicPartition)) {
+                acknowledged = false;
+            } else {
+                long commitOffset = record.offset() + 1;
+                OffsetAndMetadata offsetAndMetadata = new OffsetAndMetadata(commitOffset, "Commit From Forklift Server");
+                pendingOffsets.merge(topicPartition, offsetAndMetadata, this::greaterOffset);
```
Kuroshii: +1, that's pretty neat

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+     * @return true if the record has been acknowledged and may be processed, else false
+     * @throws InterruptedException if interrupted
+     */
+    public boolean acknowledgeRecord(ConsumerRecord<?, ?> record) throws InterruptedException {
+        lock.readLock().lock();
+        try {
+            boolean acknowledged;
+            TopicPartition topicPartition = new TopicPartition(record.topic(), record.partition());
+            if (!assignment.contains(topicPartition)) {
+                acknowledged = false;
```
Kuroshii: Can't you just `return false` and `return true` instead of setting an intermediate variable?

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Maintains a batch of acknowledged records and provides management for adding and removing {@link org.apache.kafka.common.TopicPartition partitions}.
+ * Acknowledged records are those that have started processing but have not yet been committed to the Kafka Broker.  This class is threadsafe.
+ */
+public class AcknowledgedRecordHandler {
+    private static final Logger log = LoggerFactory.getLogger(AcknowledgedRecordHandler.class);
+    private ConcurrentHashMap<TopicPartition, OffsetAndMetadata> pendingOffsets = new ConcurrentHashMap<>();
+    private Set<TopicPartition> assignment = ConcurrentHashMap.newKeySet();
+    private ReadWriteLock lock = new ReentrantReadWriteLock();
```
Kuroshii: I find your use of `ReadWriteLock` a little confusing -- I guess you're just using it to protect the non-atomic operations of `removePartitions`?

If that's the case, a little documentation of that among all of this documentation would be nice.

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -0,0 +1,349 @@
...
+    /**
+     * Stops the controller.
+     *
+     * @param timeout  the time to wait for the service to stop
+     * @param timeUnit the TimeUnit of timeout
+     * @throws InterruptedException if the thread is interrupted
+     */
+    public void stop(long timeout, TimeUnit timeUnit) throws InterruptedException {
+        executor.shutdownNow();
+        kafkaConsumer.wakeup();
```
Kuroshii: Maybe try to `wakeup` the consumer before you try to interrupt it's thread? I would imagine that that would tend to deal with stopping more elegantly.

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -0,0 +1,349 @@
...
+        running = false;
+        executor.awaitTermination(timeout, timeUnit);
```
Kuroshii: This has a result value indicating whether awaiting termination succeeded - maybe we want to log an error or throw an exception when the unexpected happens.

I would expect that this almost always succeeds, so wouldn't we want to know when it doesn't?

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -0,0 +1,349 @@
...
+        records.partitions().forEach(tp -> messages.put(tp, new ArrayList<>()));
+        for (ConsumerRecord<?, ?> record : records) {
+            TopicPartition partition = new TopicPartition(record.topic(), record.partition());
+            messages.get(partition).add(new KafkaMessage(this, record));
+        }
+        return messages;
+    }
+
+    private class RebalanceListener implements ConsumerRebalanceListener {
+        @Override public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
```
Kuroshii: A little weird to inline the `@Override` here and not elsewhere in the same class (or really anywhere else).

core/src/main/java/forklift/connectors/ForkliftMessage.java
```diff
@@ -54,25 +56,24 @@ public void setFlagged(boolean flagged) {
...
 
-    public void setProperties(Map<String, Object> properties) {
-        // Get rid of unmodifiable.
-        final Map<String, Object> newProps = new HashMap<>();
-        properties.keySet().stream().forEach(key -> newProps.put(key, properties.get(key)));
-        this.properties = newProps;
+    public void setHeaders(Map<Header, Object> headers) {
+         final Map<Header, Object> newHeaders = new HashMap<>();
+         headers.keySet().stream().forEach(key -> newHeaders.put(key, headers.get(key)));
+         this.headers = newHeaders;
```
Kuroshii: Nit: whitespace at the end of this line

core/src/main/java/forklift/consumer/Consumer.java
```diff
@@ -227,12 +239,13 @@ public void messageLoop(ForkliftConsumerI consumer) {
             running.set(true);
 
             while (running.get()) {
-                Message jmsMsg;
-                while ((jmsMsg = consumer.receive(2500)) != null && running.get()) {
-                    final ForkliftMessage msg = connector.jmsToForklift(jmsMsg);
+                ForkliftMessage consumedMsg;
```
Kuroshii: `consumedMsg` is kind of a bad name. Maybe `consumerMsg`? :man_shrugging:

core/src/main/java/forklift/consumer/Consumer.java
```diff
@@ -77,19 +73,20 @@
     private BlockingQueue<Runnable> blockQueue;
     private ThreadPoolExecutor threadPool;
 
-    private Callback<Consumer> outOfMessages;
+    private java.util.function.Consumer<Consumer> outOfMessages;
```
Kuroshii: Just an opinion: I don't necessarily agree that we should use the Java 8 `Consumer` interface  _everywhere_ it is applicable. I think it's kinda overloaded in this context and a good type name surrounding it's function is almost certainly easier to read.

But `Callback` sucks even more, and it's not really worth it to factor out all of it at this point, so yeah.

core/src/main/java/forklift/producers/ForkliftSyncProducer.java
```diff
@@ -78,13 +85,13 @@ public void setHeaders(Map<Header, Object> headers)
...
     @Override
-    public Map<String, Object> getProperties()
+    public Map<String, String> getProperties()
       throws ProducerException {
         return producer.getProperties();
     }
 
     @Override
-    public void setProperties(Map<String , Object> properties)
+    public void setProperties(Map<String , String> properties)
```
Kuroshii: History: that space still looks terrible

core/src/test/java/forklift/consumer/MessageRunnableTest.java
```diff
@@ -44,33 +70,37 @@ else if (m.isAnnotationPresent(OnValidate.class))
 
         TestConsumer1 tc = new TestConsumer1();
 
-        MessageRunnable mr = new MessageRunnable(null, new ForkliftMessage(jmsMsg), null, tc, onMessage, onValidate, null, createOnProcessStepMap(), Collections.emptyList());
+        MessageRunnable
```
Kuroshii: Not really sure of the point of putting the variable type and identifier on different lines when there is enough room?

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -0,0 +1,349 @@
...
+        executor.shutdownNow();
+        kafkaConsumer.wakeup();
+        running = false;
+        executor.awaitTermination(timeout, timeUnit);
+    }
+
+    private void controlLoop() {
+        try {
+            while (running) {
+                boolean updatedAssignment = false;
```
Kuroshii: Looks like `updatedAssignment = topicsChanged` as it's used currently?

In theory it's unnecessary, but at least it can be used more clearly?

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -0,0 +1,349 @@
...
+            } catch (InterruptedException e) {
+                log.debug("controlLoop partition rebalance interrupted", e);
+                Thread.currentThread().interrupt();
+            }
+        }
+
+        @Override
+        public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
+            log.debug("controlLoop partitions assigned");
+            acknowlegmentHandler.addPartitions(partitions);
```
Kuroshii: This block is the same as `updateAssignment`; can you just call that?

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+                acknowledged = true;
+            }
+            return acknowledged;
+        } finally {
+            lock.readLock().unlock();
+        }
+    }
+
+    private OffsetAndMetadata greaterOffset(OffsetAndMetadata a, OffsetAndMetadata b) {
+        if (a == null) {
```
Kuroshii: You shouldn't _need_ to check for null here, since the merge function will only be called if there is an existing value (which should always be non-null).

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -0,0 +1,349 @@
...
+        } else {
+            return kafkaConsumer.poll(100);
+        }
+    }
+
+    private void updateAssignment() {
+        this.acknowlegmentHandler.addPartitions(kafkaConsumer.assignment());
+        for (TopicPartition partition : kafkaConsumer.assignment()) {
+            this.flowControl.merge(partition, new AtomicInteger(),
+                                   (oldValue, newValue) -> oldValue == null ? newValue : oldValue);
```
Kuroshii: `oldValue` should never be null; otherwise the merge function wouldn't be called.

So, just use a `putIfAbsent`?

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -0,0 +1,349 @@
...
+        } catch (InterruptedException e) {
+            log.info("Interrupted, controlLoop exiting");
+        } catch (Throwable t) {
+            log.error("controlLoop error, exiting", t);
+            throw t;
+        } finally {
+            running = false;
+            try {
+                Map<TopicPartition, OffsetAndMetadata>
+                                offsetData =
```
Kuroshii: `offsetData` is a bad name

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * Maintains a batch of acknowledged records and provides management for adding and removing {@link org.apache.kafka.common.TopicPartition partitions}.
+ * Acknowledged records are those that have started processing but have not yet been committed to the Kafka Broker.  This class is threadsafe.
+ */
+public class AcknowledgedRecordHandler {
+    private static final Logger log = LoggerFactory.getLogger(AcknowledgedRecordHandler.class);
+    private ConcurrentHashMap<TopicPartition, OffsetAndMetadata> pendingOffsets = new ConcurrentHashMap<>();
```
Kuroshii: `pendingOffsets` is kinda bad name; pending what? Consumption maybe? But in terms of this class, "pending" is more or less the final state.

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -0,0 +1,349 @@
...
+                                this.acknowlegmentHandler.removePartitions(kafkaConsumer.assignment());
+                try {
+                    if (failedOffset != null) {
+                        log.debug("failedOffset of size: " + failedOffset.size());
+                        for (Map.Entry<TopicPartition, OffsetAndMetadata> entry : failedOffset.entrySet()) {
+                            offsetData.merge(entry.getKey(), entry.getValue(), (oldO, newO) -> {
```
Kuroshii: Because all of the offsets from the `acknowledgementHandler` are increasing values, can't you just `putAll` the removed offsets (`offsetData`) on top of the `failedOffsets`?

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -0,0 +1,349 @@
...
+     * Stops the controller.
+     *
+     * @param timeout  the time to wait for the service to stop
+     * @param timeUnit the TimeUnit of timeout
+     * @throws InterruptedException if the thread is interrupted
+     */
+    public void stop(long timeout, TimeUnit timeUnit) throws InterruptedException {
+        executor.shutdownNow();
+        kafkaConsumer.wakeup();
+        running = false;
```
Kuroshii: Maybe setting `running = false` so soon is a bit misleading? :man_shrugging:

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -0,0 +1,349 @@
...
+                            });
+                        }
+                    }
+
+                    log.info("closing offset size committed: " + offsetData.size());
+                    try {
+                        //if we got here through running = false or interrupt instead of a wakeup, we need
+                        //to retry our commit as the first attempt will throw a WakeupException
+                        kafkaConsumer.commitSync(offsetData);
+                    } catch (WakeupException wakeup) {
```
Kuroshii: If this exception is expected, name the exception `expected`?

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -0,0 +1,349 @@
...
+public class KafkaController {
+
+    private static final Logger log = LoggerFactory.getLogger(KafkaController.class);
+    private volatile boolean running = false;
+    private final Set<String> topics = ConcurrentHashMap.newKeySet();
+    private final ExecutorService executor = Executors.newSingleThreadExecutor();
+    private final KafkaConsumer<?, ?> kafkaConsumer;
+    private final MessageStream messageStream;
+    private volatile boolean topicsChanged = false;
+    private AcknowledgedRecordHandler acknowlegmentHandler = new AcknowledgedRecordHandler();
```
Kuroshii: It's spelled "acknowledgment" :1st_place_medal:

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,257 @@
...
+
+    @Override
+    public String send(Map<String, String> message) throws ProducerException {
+        return sendForkliftWrappedMessage(this.formatMap(message), null);
+    }
+
+    @Override
+    public String send(Map<Header, Object> headers, Map<String, String> properties, ForkliftMessage message)
+                    throws ProducerException {
+        throw new UnsupportedOperationException("Kafka Producer does not support headers or properties");
```
Kuroshii: "or properties". Well.

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,257 @@
...
+public class KafkaForkliftProducer implements ForkliftProducerI {
+
+    public final static String SCHEMA_FIELD_NAME_VALUE = "forkliftValue";
+    public final static String SCHEMA_FIELD_NAME_PROPERTIES = "forkliftProperties";
+    private final String topic;
+    private final KafkaProducer<?, ?> kafkaProducer;
+    private static final ObjectMapper mapper = new ObjectMapper().registerModule(new JavaTimeModule())
+                                                                 .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+
+    private Schema forkliftSchema = null;
```
Kuroshii: Since it's just set in the constructor, can't you make this `final`?

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,257 @@
...
+    private Map<Class<?>, Schema> avroSchemaCache = new ConcurrentHashMap<>();
+    private Map<String, String> properties = new HashMap<>();
+
+    public KafkaForkliftProducer(String topic, KafkaProducer<?, ?> kafkaProducer) {
+        this.kafkaProducer = kafkaProducer;
+        this.topic = topic;
+        Schema.Parser parser = new Schema.Parser();
+        this.forkliftSchema = parser.parse(
```
Kuroshii: Isn't this constant and shared across all producers? Then can't it be static?

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,257 @@
...
+
+    @Override
+    public void close() throws IOException {
+        //do nothing, the passed in KafkaProducer may be used elsewhere and should be closed by the KafkaController
+    }
+
+    private String sendForkliftWrappedMessage(String message, Map<String, String> messageProperties) throws ProducerException {
+        GenericRecord avroRecord = new GenericData.Record(forkliftSchema);
+        avroRecord.put(SCHEMA_FIELD_NAME_VALUE, message);
+        messageProperties = messageProperties == null ? new HashMap<>() : new HashMap<>(messageProperties);
```
Kuroshii: The logic here seems backwards to me. I would think to start with a copy of `properties` and then `putAll` the `messageProperties` on top of it.

Then you can do something like:
```
Map<String, String> appliedProperties = new HashMap<>(properties);

if (messageProperties == null)
    messageProperties = Collections.emptyMap();

appliedProperties.putAll(messageProperties);
```

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,257 @@
...
+                throw new ProducerException("Error sending Kafka Message", e);
+            } catch (ExecutionException e) {
+                throw new ProducerException("Error creating Kafka Message", e);
+            }
+        } catch (IOException e) {
+            throw new ProducerException("Error creating Kafka Message", e);
+        }
+    }
+
+    private String formatMap(Map<String, String> map) {
```
Kuroshii: Won't this encoding have trouble with maps that have keys or values with `=` or a newline?

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,257 @@
...
+        //Write message to json
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        Encoder encoder = EncoderFactory.get().jsonEncoder(message.getSchema(), outputStream);
+        DatumWriter<SpecificRecord> writer = new SpecificDatumWriter<>(message.getSchema());
+        writer.write(message, encoder);
+        encoder.flush();
+        String json = new String(outputStream.toByteArray(), Charset.forName("UTF-8"));
+
+        //modify schema to include forklift properties
+        Schema modifiedSchema = avroSchemaCache.get(message.getClass());
```
Kuroshii: I don't like this cache, but would it hurt to make it static?

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
Kuroshii: If it's a one-time read, maybe it could just be in it's own file as well.

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,257 @@
...
+                                                   "\"doc\":\"Properties added to support forklift interfaces. Format is key,value entries delimited with new lines\"}");
+        ObjectNode schemaNode = (ObjectNode)mapper.readTree(originalJson);
+        ArrayNode fieldsNode = (ArrayNode)schemaNode.get("fields");
+        fieldsNode.add(propertiesField);
+        schemaNode.set("fields", fieldsNode);
+        Schema.Parser parser = new Schema.Parser();
+        return parser.parse(mapper.writeValueAsString(schemaNode));
+    }
+
+    private GenericRecord addForkliftPropertiesToAvroObject(SpecificRecord message) throws IOException {
```
Kuroshii: A lot of what is going on here seems pretty inefficient. What's worse, this is done for every avro message we send out. Producing events is supposed to be fast.

I bet that the vast majority of our (future) use cases for a producer don't even use properties; could we partially optimize by not doing any of the hard work if we have no properties to insert?

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,257 @@
...
+        avroRecord.put(SCHEMA_FIELD_NAME_VALUE, message);
+        messageProperties = messageProperties == null ? new HashMap<>() : new HashMap<>(messageProperties);
+        //add the producer level properties but do not overwrite message level properties
+        for (Map.Entry<String, String> entry : this.properties.entrySet()) {
+            messageProperties.putIfAbsent(entry.getKey(), entry.getValue());
+        }
+        avroRecord.put(SCHEMA_FIELD_NAME_PROPERTIES, this.formatMap(messageProperties));
+        ProducerRecord record = new ProducerRecord<>(topic, null, avroRecord);
+        try {
+            RecordMetadata result = (RecordMetadata)kafkaProducer.send(record).get();
```
Kuroshii: Can we stay asynchronous? Batching messages together is one of the core things that makes Kafka fast, so sending them synchronously one-at-a-time is missing out on a lot.

If you don't want the client of this code to deal with `RecordMetadata`, there are better ways than just making this synchronous (another callback, returning a different future, etc).

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,257 @@
...
+            return result.topic() + "-" + result.partition() + "-" + result.offset();
+
```
Kuroshii: Nit: this newline looks weird.

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,257 @@
...
+            GenericRecord avroRecord = addForkliftPropertiesToAvroObject(message);
+            ProducerRecord record = new ProducerRecord<String, GenericRecord>(topic, null, avroRecord);
+            try {
+                RecordMetadata result = (RecordMetadata)kafkaProducer.send(record).get();
+                return result.topic() + "-" + result.partition() + "-" + result.offset();
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new ProducerException("Error sending Kafka Message", e);
+            } catch (ExecutionException e) {
+                throw new ProducerException("Error creating Kafka Message", e);
```
Kuroshii: This exception message doesn't make sense; the message is completely created before the corresponding `try` block.

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,257 @@
...
+ *               "doc":"Properties added to support forklift interfaces. Format is key,value entries delimited with new lines"}]}
+ * </pre>
+ * <p>
+ * Headers are not supported and calls to the {@link #send(java.util.Map, java.util.Map, forklift.connectors.ForkliftMessage)}
+ * or {@link #setHeaders(java.util.Map)} will result in an {@link java.lang.UnsupportedOperationException}.
+ */
+public class KafkaForkliftProducer implements ForkliftProducerI {
+
+    public final static String SCHEMA_FIELD_NAME_VALUE = "forkliftValue";
+    public final static String SCHEMA_FIELD_NAME_PROPERTIES = "forkliftProperties";
```
Kuroshii: These constant names seem weird to me.

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,257 @@
...
+ * Implementation of the {@link forklift.producers.ForkliftProducerI}.  Messages sent are fully integrated
+ * with confluent's schema-registry.  Avro compiled objects may be sent through the {@link #send(Object)} method.  If
+ * an avro object is sent, the schema will be evolved to include the {@link #SCHEMA_FIELD_NAME_PROPERTIES} field as
+ * follows:
+ * <pre>
+ * {"name":"forkliftProperties","type":"string","default":"",
+ *  "doc":"Properties added to support forklift interfaces. Format is key,value entries delimited with new lines"}
+ * </pre>
+ * <p>
+ * The value of the forkliftProperties will be key,value entries delimited with a newline
```
Kuroshii: This isn't what you're doing.

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -0,0 +1,257 @@
...
+            Thread.currentThread().interrupt();
+            throw new ProducerException("Error sending Kafka Message", e);
+        } catch (ExecutionException e) {
+            throw new ProducerException("Error sending Kafka Message", e);
+        }
+    }
+
+    private Schema addForkliftPropertiesToSchema(Schema schema) throws IOException {
+        String originalJson = schema.toString(false);
+        JsonNode propertiesField = mapper.readTree("{\"name\":\"forkliftProperties\",\"type\":\"string\",\"default\":\"\"," +
```
Kuroshii: This inline JSON seems hard to read; can this at least be factored into a constant?

connectors/kafka/src/main/java/forklift/message/KafkaMessage.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    private final void parseRecord() {
```
Kuroshii: Nit: This method could use some more spacing.

connectors/kafka/src/main/java/forklift/message/KafkaMessage.java
```diff
@@ -0,0 +1,67 @@
...
+            }
+            value = genericRecord.get(KafkaForkliftProducer.SCHEMA_FIELD_NAME_VALUE);
+            //If the value is null, this is most likely an avro object
+            if (value == null) {
+                String jsonValue = genericRecord.toString();
+                value = jsonValue != null && jsonValue.startsWith("{") ? jsonValue : null;
+            }
+        }
+        if (value == null) {
+            this.setFlagged(true);
```
Kuroshii: What was the point of exposing the variables in `ForkliftMessage` as `protected` if you're just using the setters and getters directly anyways?

connectors/kafka/src/main/java/forklift/message/KafkaMessage.java
```diff
@@ -0,0 +1,67 @@
...
+    public String getId() {
+        return consumerRecord.topic() + "-" + consumerRecord.partition() + "-" + consumerRecord.offset();
+    }
+
+    /**
+     * <strong>WARNING:</strong> Called from constructor
+     */
+    private final void parseRecord() {
+        Object value = null;
+        if (consumerRecord.value() instanceof GenericRecord) {
```
Kuroshii: This block could basically be a logically distinct method from the forklift message handling piece.

connectors/kafka/src/main/java/forklift/message/KafkaMessage.java
```diff
@@ -0,0 +1,67 @@
...
+    public boolean acknowledge() throws ConnectorException {
+        try {
+            return controller.acknowledge(consumerRecord);
+        } catch (InterruptedException e) {
+            throw new ConnectorException("Error acknowledging message");
+        }
+    }
+
+    @Override
+    public String getId() {
```
Kuroshii: This is a weird: there's still an internal field for `id`, and you can set it, but what you `get` isn't what you `set`.

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -0,0 +1,349 @@
...
+            messages.get(partition).add(new KafkaMessage(this, record));
+        }
+        return messages;
+    }
+
+    private class RebalanceListener implements ConsumerRebalanceListener {
+        @Override public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
+            log.debug("controlLoop partitions revoked");
+            try {
+                Map<TopicPartition, OffsetAndMetadata>
```
Kuroshii: Still don't get this convention.

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -0,0 +1,349 @@
...
+                                                 .incrementAndGet());
+            messageStream.addRecords(consumerRecordsToKafkaMessages(records));
+        }
+    }
+
+    private void commitOffsets(Map<TopicPartition, OffsetAndMetadata> offsets) {
+        if (offsets.size() > 0) {
+            //if the commit fails, perhaps due to a wakeup, store the offsets so they can be retried before the controller stops
+            failedOffset = offsets;
+            kafkaConsumer.commitSync(offsets);
```
Kuroshii: Why is everything a `commitSync`? Isn't `commitAsync` faster and just as reliable?

connectors/activemq/src/main/java/forklift/producers/ActiveMQProducer.java
Kuroshii: Got it

connectors/activemq/src/main/java/forklift/producers/ActiveMQProducer.java
Kuroshii: Got it

connectors/activemq/src/main/java/forklift/producers/ActiveMQProducer.java
Kuroshii: Got it

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
Kuroshii: Got it

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
AFrieze: commitAsync is faster, but it can error out in the background.  Since we are asynchronously processing workload in the background already, this seemed like an ok approach but I'm open to changing it.  Seems like something that can be left as a future optimization if needed.

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
Kuroshii: Got it

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
AFrieze: The spacing?  Not intentional, most likely my auto-formatter.  I'll fix.

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
Kuroshii: Got it

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
Kuroshii: Got it

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
Kuroshii: Got it

connectors/kafka/src/main/java/forklift/message/KafkaMessage.java
AFrieze: Agreed, it looks like @dcshock added the setter to support activeMQ directly using the ForkliftMessage.  I tend to think the ForkliftMessage should be an interface or abstract class without setId or setMsg.  Not sure we will get a fix in for this one on this pass through.

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
Kuroshii: "Fixing"

plugins/retry/src/main/java/forklift/retry/RetryHandler.java
Kuroshii: Got it

connectors/kafka/src/main/java/forklift/message/KafkaMessage.java
AFrieze: I'll take a look

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
AFrieze: I'll take a look

connectors/kafka/src/main/java/forklift/message/MessageStream.java
Kuroshii: Got it

connectors/kafka/src/main/java/forklift/message/MessageStream.java
Kuroshii: Got it

connectors/kafka/src/main/java/forklift/message/MessageStream.java
Kuroshii: Got it

connectors/kafka/src/main/java/forklift/message/MessageStream.java
Kuroshii: Got it; just going to log

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
AFrieze: good catch (facepalm)

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
AFrieze: changed to sending

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
AFrieze: removed

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
AFrieze: I agree and my thought was to return a future which I think was briefly discussed in our last meeting.  The big problem is that we are implementing an interface that returns a String.  I suggest we take the performance hit for the ease of integration, knowing that we can come back and change the interface or implement an asyn connector that doesn't return id.

connectors/kafka/src/main/java/forklift/message/KafkaMessage.java
Kuroshii: Couldn't it just be fixed by calling `setId` in the constructor?

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
Kuroshii: Got it

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
Kuroshii: Got it

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
Kuroshii: Got it.

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
AFrieze: Good idea, I'll get this implemented.  There are also inefficiencies on the reading side that could potentially be improved with changes to the core in the future.

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
Kuroshii: I'll take care of these constants

connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
Kuroshii: Got it

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
AFrieze: I'll change it to static.  What do you know like about it?

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
AFrieze: I would think so, but it's the pre-existing format that forklift uses.

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
AFrieze: I'm fine with this.  Will refactor

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
AFrieze: Got it

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
Kuroshii: I don't like it because it exists to help optimize out the cost of doing something that doesn't seem like it should really be that expensive.

However, my judgment on that could be wrong.

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
AFrieze: I think this is true.  I'm adding in intermediate map called "finalOffsets" for now and running some tests.

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
AFrieze: I'll rename to simply "offsets" unless you have another suggestion

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
AFrieze: Agreed, I'll change to just offsets for now

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
Kuroshii: FeelsBadMan

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
Kuroshii: `offsets` is okay, `acknowledgedOffsets` is a little better, but whatever :man_shrugging:

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
AFrieze: Got it

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
AFrieze: Got it

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
AFrieze: I'll add a comment

connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
AFrieze: Got it

connectors/kafka/README.md
AFrieze: Yes, reworded a bit

connectors/kafka/src/schemas/Test.avsc
Kuroshii: Got it

Kuroshii: Got it

core/src/main/java/forklift/connectors/ForkliftMessage.java
Kuroshii: Got it

core/src/main/java/forklift/consumer/Consumer.java
Kuroshii: Got it

core/src/main/java/forklift/producers/ForkliftSyncProducer.java
Kuroshii: Got it

plugins/retry/src/main/java/forklift/retry/RetryES.java
Kuroshii: Got it

connectors/kafka/src/main/java/forklift/message/KafkaMessage.java
AFrieze: Not really.  You can't set an id on a kafka message, we would need to wire it into the properties or something.  I think the responsibilities of the ForkliftMessage class are worth a discussion at some point.

connectors/kafka/src/main/java/forklift/message/KafkaMessage.java
AFrieze: Mainly as a warning to other developer.  I suppose marking it final serves as enough warning...I just in general dislike calling methods from the constructor

PR #126, opened 2017-04-19
Feature/constructor level injection

AFrieze:
Added constructor injection for consumers.

Note:  I thought about pulling this functionality out into its own set of classes, or even incorporating an existing DI framework.  I decided that was a bit more than I wanted to take on right now and kept code changes to a minimum.  Could be considerations for the future though.


core/classpath.txt
```diff
@@ -0,0 +1 @@
+/Applications/activator-dist-1.3.7/repository/com.google.guava/guava/18.0/bundles/guava.jar:/Users/afrieze/.ivy2/cache/com.fasterxml.jackson.core/jackson-databind/bundles/jackson-databind-2.7.3.jar:/Users/afrieze/.ivy2/cache/com.fasterxml.jackson.core/jackson-annotations/bundles/jackson-annotations-2.7.0.jar:/Users/afrieze/.ivy2/cache/com.fasterxml.jackson.core/jackson-core/bundles/jackson-core-2.7.3.jar:/Users/afrieze/.ivy2/cache/com.fasterxml.jackson.datatype/jackson-datatype-jsr310/bundles/jackson-datatype-jsr310-2.7.3.jar:/Users/afrieze/.ivy2/cache/ch.qos.logback/logback-classic/jars/logback-classic-1.0.13.jar:/Users/afrieze/.ivy2/cache/ch.qos.logback/logback-core/jars/logback-core-1.0.13.jar:/Users/afrieze/.ivy2/cache/org.slf4j/slf4j-api/jars/slf4j-api-1.7.5.jar:/Users/afrieze/.ivy2/cache/org.reflections/reflections/jars/reflections-0.9.10.jar:/Applications/activator-dist-1.3.7/repository/org.javassist/javassist/3.18.2-GA/bundles/javassist.jar:/Users/afrieze/.ivy2/cache/com.google.code.findbugs/annotations/jars/annotations-2.0.1.jar:/Applications/activator-dist-1.3.7/repository/javax.inject/javax.inject/1/jars/javax.inject.jar
```
Kuroshii: This should be ignored or deleted, right?

core/src/test/java/forklift/consumer/injection/ConsumerTest.java
```diff
@@ -127,6 +134,63 @@ public void injectJson() {
...
+    /**
+     * Tests constructor injection into a consumer.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void jsonConstructorInjection() throws Exception {
+        ConsumerService service = new ConsumerService(ServiceBeanResolver.class);
+        Consumer test = new Consumer(ConstructorJsonConsumer.class, forklift, this.getClass().getClassLoader());
+        test.setServices(Arrays.asList(service));
```
Kuroshii: Maybe a little cleaner as `addServices`

core/src/test/java/forklift/consumer/injection/ConsumerTest.java
```diff
@@ -1,15 +1,18 @@
-package forklift;
+package forklift.consumer.injection;
```
Kuroshii: Hey, it's not all injection tests!

core/src/test/java/forklift/consumer/injection/ConsumerTest.java
Kuroshii: Minor: Also, maybe it'd be nice to test service-based injection separately from forklift-based injection

core/src/main/java/forklift/consumer/Consumer.java
```diff
@@ -213,7 +222,7 @@ else if (queue != null)
                 log.info("Creating thread pool of {}", multiThreaded.value());
                 blockQueue = new ArrayBlockingQueue<>(multiThreaded.value() * 100 + 100);
                 threadPool = new ThreadPoolExecutor(
-                    multiThreaded.value(), multiThreaded.value(), 5L, TimeUnit.MINUTES, blockQueue);
+                                                       multiThreaded.value(), multiThreaded.value(), 5L, TimeUnit.MINUTES, blockQueue);
```
Kuroshii: That indentation seems a bit excessive.

core/src/main/java/forklift/consumer/Consumer.java
```diff
@@ -344,10 +354,22 @@ public void shutdown() {
     public void setOutOfMessages(java.util.function.Consumer<Consumer> outOfMessages) {
         this.outOfMessages = outOfMessages;
     }
+    
+    private final void configureConstructorInjection() {
+        Constructor<?>[] constructors = msgHandler.getDeclaredConstructors();
+        for (Constructor<?> constructor : constructors) {
+            if (constructor.isAnnotationPresent(Inject.class)) {
+                this.constructor = constructor;
```
Kuroshii: Maybe worth it to log an error or warning (realistically, logging an error is more immediately useful) when there are multiple constructors annotated with `@Inject`?

core/src/main/java/forklift/consumer/Consumer.java
```diff
@@ -483,10 +405,154 @@ else if (producer.topic().length() > 0)
...
+            instance = msgHandler.newInstance();
+        }
+        return instance;
+    }
+
+    private Object[] buildConstructorParameters(ForkliftMessage forkliftMessage, List<Closeable> closeables) throws IOException {
+        Object[] parameters = new Object[constructorAnnotations.length];
+        int index = 0;
+        for (Annotation[] parameterAnnotations : constructorAnnotations) {
+            //there must be at least one annotation that indicates what we should inject
```
Kuroshii: Except not, right? If there's no annotation you still handle it fine by treating it like it has `Inject`

core/src/main/java/forklift/consumer/Consumer.java
```diff
@@ -483,10 +405,154 @@ else if (producer.topic().length() > 0)
...
+    }
+
+    private Object getInjectableValue(Annotation decorator, String mappedName, Class<?> mappedClass, ForkliftMessage msg) throws IOException {
+        Object value = null;
+        if (decorator == null || decorator.annotationType() == javax.inject.Inject.class) {
+            if (this.services != null) {
+                // Try to resolve the class from any available BeanResolvers.
+                for (ConsumerService s : this.services) {
+                    try {
+                        final Object o = s.resolve(mappedClass, null);
```
Kuroshii: So we don't want name-based resolution? I guess it's worked so far, but isn't it a bit limiting to be unable to distinguish objects of the same type?

core/src/main/java/forklift/consumer/Consumer.java
```diff
@@ -483,10 +405,154 @@ else if (producer.topic().length() > 0)
...
+                        if (o != null) {
+                            value = o;
+                            break;
+                        }
+                    } catch (Exception e) {
+                        log.debug("", e);
```
Kuroshii: I know this was existing code, but aren't errors like "failed to autowire dependency" pretty useful?

core/src/main/java/forklift/consumer/Consumer.java
```diff
@@ -483,10 +405,154 @@ else if (producer.topic().length() > 0)
...
+        }
+        return value;
+    }
+
     public Class<?> getMsgHandler() {
         return msgHandler;
     }
 
+    /**
+     * Creates an instance of hte MessageHandler class utilized by this constructor.  Constructor and Field level injection is performed using both the
```
Kuroshii: "hte"

core/src/main/java/forklift/consumer/Consumer.java
```diff
@@ -242,11 +251,12 @@ public void messageLoop(ForkliftConsumerI consumer) {
                 ForkliftMessage consumerMsg;
                 while ((consumerMsg = consumer.receive(2500)) != null && running.get()) {
                     try {
-                        final Object handler = msgHandler.newInstance();
+
```
Kuroshii: Not really a fan of newlines at the start of a block

core/src/main/java/forklift/consumer/Consumer.java
```diff
@@ -344,10 +354,22 @@ public void shutdown() {
...
+                break;
+            }
+        }
+    }
 
     /**
      * Inject the data from a forklift message into an instance of the msgHandler class.
-     * @param msg containing data
+     *
+     * @param msg      containing data
```
Kuroshii: Doesn't really improve the vague documentation to add some spaces

core/src/main/java/forklift/consumer/Consumer.java
```diff
@@ -483,10 +405,154 @@ else if (producer.topic().length() > 0)
...
+            if (value != null && value instanceof ForkliftProducerI) {
+                closeables.add((ForkliftProducerI)value);
+            }
+
+            index++;
+        }
+        return parameters;
+    }
+
+    private Object getInjectableValue(Annotation decorator, String mappedName, Class<?> mappedClass, ForkliftMessage msg) throws IOException {
```
Kuroshii: Thought for later: maybe this could be separated out better with a map of handlers or something

core/src/main/java/forklift/consumer/Consumer.java
```diff
@@ -362,112 +384,12 @@ public void setOutOfMessages(java.util.function.Consumer<Consumer> outOfMessages
                 fields.get(clazz).forEach(f -> {
```
Kuroshii: Can we take this opportunity to not call this variable `f`?

core/src/main/java/forklift/consumer/Consumer.java
```diff
@@ -483,10 +405,154 @@ else if (producer.topic().length() > 0)
...
         return msgHandler;
     }
 
+    /**
+     * Creates an instance of hte MessageHandler class utilized by this constructor.  Constructor and Field level injection is performed using both the
+     * passed in msg and any Services {@link #addServices(ConsumerService...) added} to this consumer.
+     *
+     * @param msg
+     */
+    public Object getMsgHandlerInstance(ForkliftMessage msg) throws InvocationTargetException, IOException, InstantiationException, IllegalAccessException {
```
Kuroshii: Is this method just public for internal testing, or what?

Kuroshii: I can't imagine most callers to these methods having a meaningful way of handling all of these exceptions differently. Maybe just wrap them in one exception?

core/src/test/java/forklift/consumer/injection/ConsumerTest.java
AFrieze: I think it's worth considering a refactor where the injection code is removed from the consumer entirely(this is what I started to do initially).  

The setServices method was already in place and I tried to stick with adding the new functionality and not go down a refactoring rabbit hole.

core/src/main/java/forklift/consumer/Consumer.java
AFrieze: +1

core/src/main/java/forklift/consumer/Consumer.java
AFrieze: +1

core/src/main/java/forklift/consumer/Consumer.java
AFrieze: Yes.  Copying how the inject method and testing works.

AFrieze: Good catch.  The "Inject" method throws a runtime exception so I'l do the same

core/src/main/java/forklift/consumer/Consumer.java
AFrieze: I'm avoiding making changes to the existing functionality.

core/src/main/java/forklift/consumer/Consumer.java
AFrieze: the space is to align with the param below that isn't shown :(

PR #127, opened 2017-04-19
Fixed bug with adding a removed topic

AFrieze:
This fixes a bug with adding a topic which was previously removed.

### DIFF
--- a/connectors/kafka/src/main/java/forklift/controller/KafkaController.java
+++ b/connectors/kafka/src/main/java/forklift/controller/KafkaController.java
@@ -12,6 +12,7 @@ import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.errors.WakeupException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -24,6 +25,9 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+
+import javax.annotation.concurrent.GuardedBy;
 
 /**
  * Manages the {@link org.apache.kafka.clients.consumer.KafkaConsumer} thread.  Polled records are sent to the MessageStream. Commits
@@ -41,10 +45,13 @@ public class KafkaController {
     private final ExecutorService executor = Executors.newSingleThreadExecutor();
     private final KafkaConsumer<?, ?> kafkaConsumer;
     private final MessageStream messageStream;
-    private volatile boolean topicsChanged = false;
     private AcknowledgedRecordHandler acknowledgmentHandler = new AcknowledgedRecordHandler();
     private Map<TopicPartition, OffsetAndMetadata> failedOffset = null;
     private Map<TopicPartition, AtomicInteger> flowControl = new ConcurrentHashMap<>();
+    @GuardedBy("this")
+    private String topicToAdd;
+    @GuardedBy("this")
+    private String topicToRemove;
 
     public KafkaController(KafkaConsumer<?, ?> kafkaConsumer, MessageStream messageStream) {
         this.kafkaConsumer = kafkaConsumer;
@@ -67,12 +74,18 @@ public class KafkaController {
      * @param topic the topic to subscribe to
      * @return true if the topic was added, false if already added
      */
-    public synchronized boolean addTopic(String topic) {
+    public synchronized boolean addTopic(String topic) throws InterruptedException {
+        while (topicToAdd != null) {
+            this.wait();
+        }
         if (!topics.contains(topic)) {
-            messageStream.addTopic(topic);
-            topics.add(topic);
-            topicsChanged = true;
+            this.topicToAdd = topic;
+            //Notify that topics have changed
             this.notifyAll();
+            while (this.topicToAdd != null && running) {
+                //wait for the control loop to add the topic
+                this.wait();
+            }
             return true;
         }
         return false;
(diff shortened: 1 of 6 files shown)


connectors/kafka/src/test/java/forklift/integration/ConsumerActionTests.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class ConsumerActionTests extends BaseIntegrationTest {
+
+    /**
+     * Test subscribing to a topic, unsubscribing, then resubscribing to the topic.
+     */
+    @Test
+    public void topicTurnoverTest() throws StartupException, ConnectorException, InterruptedException, ProducerException {
+
+        String topic = "forklift-string-topic";
```
Kuroshii: You declare this "constant" but don't use it

connectors/kafka/src/test/java/forklift/connectors/KafkaControllerTests.java
```diff
@@ -52,14 +53,27 @@ public void teardown() throws InterruptedException {
...
     @Test
-    public void addTopicFalseTest() {
+    public void addingTheSameTopicRepeatedlyDoesntLock() throws InterruptedException {
+
+        this.controller.start();
+        String topic1 = "topic1";
+        for(int i = 0; i < 10; i++){
+            this.controller.addTopic(topic1);
+        }
+        assertTrue(true);
```
Kuroshii: Seems a bit redundant; if you want to clarify this test, an comment would be more expressive

connectors/kafka/src/test/java/forklift/connectors/KafkaControllerTests.java
```diff
@@ -52,14 +53,27 @@ public void teardown() throws InterruptedException {
...
         boolean added = this.controller.addTopic(topic1);
         assertEquals(true, added);
     }
 
     @Test
-    public void addTopicFalseTest() {
+    public void addingTheSameTopicRepeatedlyDoesntLock() throws InterruptedException {
+
+        this.controller.start();
+        String topic1 = "topic1";
```
Kuroshii: MInor: could be final

connectors/kafka/src/test/java/forklift/integration/ConsumerActionTests.java
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.Test;
+
+public class ConsumerActionTests extends BaseIntegrationTest {
+
+    /**
+     * Test subscribing to a topic, unsubscribing, then resubscribing to the topic.
+     */
+    @Test
+    public void topicTurnoverTest() throws StartupException, ConnectorException, InterruptedException, ProducerException {
+
```
Kuroshii: Your use of newlines like this seems inconsistent in this PR; I personally don't like this style, but :man_shrugging:

connectors/kafka/src/test/java/forklift/integration/RebalanceTests.java
```diff
@@ -18,7 +18,7 @@
  * Tests which focus on causing partitions to be rebalanced.
  */
 public class RebalanceTests extends BaseIntegrationTest {
-
+
```
Kuroshii: Nit: indentation

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -212,39 +238,59 @@ private void controlLoop() {
...
+            return true;
+        }
+        return false;
+    }
+
+    private boolean processTopicRemoved() throws InterruptedException {
+        if (topicToRemove != null) {
+            topics.remove(topicToRemove);
+            messageStream.removeTopic(topicToRemove);
+            Set<TopicPartition> removed = kafkaConsumer.assignment().stream().filter(partition -> !topics.contains(partition.topic())).collect(Collectors.toSet());
```
Kuroshii: Is it not equivalent to check `partition.topic() == topicToRemove` instead of `!topics.contains(partition.topic())` here?

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -212,39 +238,59 @@ private void controlLoop() {
...
 
-    private void commitPendingAndWaitForTopics() throws InterruptedException {
-        //check if the last remaining topic was removed
-        if (kafkaConsumer.assignment().size() > 0) {
-            Map<TopicPartition, OffsetAndMetadata>
-                            offsets =
-                            acknowledgmentHandler.removePartitions(kafkaConsumer.assignment());
+    private boolean processTopicChanges() throws InterruptedException {
+        boolean topicsChanged = processTopicAdd();
+        topicsChanged = processTopicRemoved() || topicsChanged;
```
Kuroshii: I see that you're trying to avoid short-circuit evaluation; I think this would look clearer as two variable declarations. E.g.

```java
boolean topicsAdded = processTopicAdd();
boolean topicsRemoved = processTopicRemoved();

if (topicsAdded || topicsRemoved) {
    kafkaConsumer.subscribe(topics, new RebalanceListener());
}
```

At the very least, I think a comment is warranted for clarity, since it isn't really intuitively obvious why the code is structured this way and, the `processTopicAdd()` and `processTopicRemoved()` methods do have symmetric roles.

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
AFrieze: good catch

connectors/kafka/src/test/java/forklift/connectors/KafkaControllerTests.java
AFrieze: Good catch.  This is a bad test as if it locks you have to kill the test... I'll rework it.

AFrieze:
wip/forklift-kafka-incremental-controller removes the need for this bug fix.

PR #128, opened 2017-05-10
Add Replay and Retry Support for Kafka

Kuroshii:
#### Changes:
- Refactor how annotations like `@Queue` and `@Topic` are processed to be generalized by the idea of a "source" (`SourceI` in code). Some zealous refactoring followed from this.
- Add a `GroupedTopicSource` through the `@GroupedTopic` annotation; this represents a topic that is accessed in a queued way according to group name. This is not strictly necessary for this update, but is a nice thing to have.
- Add a `RoleInputSource` through the `@RoleInput` annotation; this is an indirect or `LogicalSource` that represents the means by which this consumer should consume messages that are intended for the function or _role_ that is performed by that consumer.
- Add a `RoleInputMessage` container that contains the information needed to create a message for a role to consume
- Add a `RoleInputConsumerWrapper`, which is a "higher-order" consumer, in the sense that it produces role messages by deserializing data from messages taken off of the consumer passed to it.
- Add support for consuming off of the new sources
- Add the concept of a `ForkliftSerializer`, which can serialize and deserialize message to be produced linked to a specific source. This is placed on the connector and is used by replay and retry to figure out the format of the messages to put in logs.
- Make replay processing idempotent through elasticsearch versioning
- Upgrade the replay and retry plugins to have more attributes:
    - the version of the plugin being used (`forklift-replay-version` or `forklift-retry-version`); this is nice for the retry / replay manager which needs to figure out _how_ to go about replaying a particular log
    - the connector to replay / retry a message to (`destination-connector`)
    - the destination to replay / retry a message to (`destination-name`)
    - the message to send to the destination (`destination-message`)
    - the format of the message to send (`destination-message-format`), either `raw-string` or `base64-bytes`
- Make the replay and retry plugin a little more modular

#### Notes:
- A `@RoleInput` annotation is now required for replay and retry, but the same source can be shared.
- Because of the generality of how `LogicalSource`s are resolved, it wouldn't be difficult to make `@RoleInput` optionally take an existing source to map back to, if we need to avoid making an extra queue or topic in particular cases.
It might look like
```java
@RoleInput(
   role = "MyApp",
   source = @Queue("myapp-queue")
)
```
- Replay and retry messages should store similar enough info that the `runRetries` flag in `RetryES` should be unnecessary eventually, as there should be enough info that retries can be restarted off of the retry-manager (`forklift-gui`).
- I kept the `getTopic` and `getQueue` methods on `ForkliftConnectorI` for compatibility, but maybe it's worth removing them since they have been generalized by `getConsumerForSource`.
- Maybe building a producer for a source should be similarly generalized in the connector (and `getTopicProducer` and `getQueueProducer` removed or deprecated accordingly).
- I disabled file-based retry because it seems difficult to support and outdated. Perhaps it should be removed altogether then.
- I had `forklift-replay-version` and `forklift-retry-version` start at `2`, so that log messages without a version can be treated as version `1`.
- There is corresponding update to `forklift-gui` to make this update to replay and retry usable: https://github.com/dcshock/forklift-gui/pull/16


connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
```diff
@@ -115,7 +115,17 @@ public KafkaForkliftProducer(String topic, KafkaProducer<?, ?> kafkaProducer) {
 
     @Override
     public String send(String message) throws ProducerException {
-        return sendForkliftWrappedMessage(message, null);
+        // let the schema registry just treat it as an avro string and ignore producer properties
+        ProducerRecord record = new ProducerRecord<>(topic, null, message);
```
AFrieze: Why ignore producer level properties?  My thought was that producer level properties always go with the message unless they are overridden  by message level.

core/src/main/java/forklift/consumer/ConsumerThread.java
```diff
@@ -34,18 +34,22 @@ public ConsumerThread(Consumer consumer) {
...
         int connectAttempt = 0;
         do {
             lastConnectAttemptTime = LocalDateTime.now();
 
-            log.info("starting consumer");
             try {
                 consumer.listen();
             } catch (Exception e) {
-                log.debug("Couldn't get connection", e);
+                if (connectAttempt == 0) {
```
AFrieze: Nice

connectors/kafka/src/main/java/forklift/producers/KafkaForkliftProducer.java
Kuroshii: Well, I wanted to write just raw role message strings to the role-input source, assuming properties weren't relevant. This may not be the right thing to do, but there are a few things to consider here:
- The schema registry will not allow us to produce fundamentally different schemas to the same topic (i.e. the `string` schema and the forklift wrapping schema in this case, so our conditional wrapping of messages is probably wrong), because they aren't compatible.
- Also, I don't really see a reason to have Forklift GUI and Forklift produce different formats for role messages.
- Producer properties feel like bloat to me, because they complicate the responsibility of `ForkliftProducerI`. And really, "producer properties" aren't even _hard_ to do on the client side.

This last bit is the key here: in my opinion, `ForkliftProducerI` has too many responsibilities:
 - It's supposed to send data.
 - It's also supposed to serialize data.
 - It's also supposed to make sure certain information is carried with the data, whether it's already serialized or not.

The last responsibility kind of made sense when AMQ was handling that for us, but it really doesn't make sense as a core part of a producer that uses Kafka.

Because of this, `ForkliftProducerI` can't be used for _just_ sending data without the other two responsibilities becoming part of the problem. Then to get what I want out of `ForkliftProducerI` (just sending), I have to dip into its responsibilities so that they can be duplicated on the GUI side.

One of a few things has to happen here:
- Forklift GUI needs to know how to perform `ForkliftProducerI`'s responsibilities (knowing how to wrap also requires knowing how to serialize) - seems too distributed.
- `ForkliftProducerI` is redesigned to have it's responsibilities be more amenable. Also seems like a lot of work.
- Something in Forklift, outside of `ForkliftProducerI`, needs to know or expose how `ForkliftProducerI` performs this wrapping so that it can be done on behalf of Forklift GUI (either through duplication, or though a more careful redesign). This seems like a lot of work, when I don't really _need_ any wrapping. It's annoying to complicate the interfaces. Add a `sendRaw` to the producer interface? Then what about AMQ which has no problem with doing that with send? Expose a "raw producer" in the connector interface? Seems ugly and already halfway to the previous solution.
- Ignore a bloated responsibility in one part of one implementation of`ForkliftProducerI` and risk breaking clients that expect it to perform all of its responsibilities all the time.

core/src/main/java/forklift/source/decorators/Queues.java
```diff
@@ -1,14 +1,17 @@
-package forklift.decorators;
+package forklift.source.decorators;
```
AFrieze: Moving these files will force a refactor on those upgrading.  Might be ok but wanted to note it.

Kuroshii: Yeah. The `forklift.decorators` namespace is getting cluttered and the source annotations have a substantially different meaning from the rest of the forklift annotations.

It's obviously not a necessary change, but I think having good code organization is important when you can get it for cheap.

core/src/main/java/forklift/consumer/wrapper/RoleInputConsumerWrapper.java
```diff
@@ -0,0 +1,27 @@
+package forklift.consumer.wrapper;
+
+import forklift.consumer.ForkliftConsumerI;
+import forklift.connectors.ConnectorException;
+import forklift.connectors.ForkliftMessage;
+
```
AFrieze: Some class level javadoc would be nice here

connectors/activemq/src/main/java/forklift/connectors/ActiveMQConnector.java
```diff
@@ -122,6 +156,13 @@ public ForkliftProducerI getTopicProducer(String name) {
     }
 
     @Override
+    public ActionSource mapSource(LogicalSource source) {
+        return source
+            .apply(RoleInputSource.class, roleSource -> new QueueSource("forklift-role-" + roleSource.getRole()))
```
AFrieze: Still working on my understanding of this all.  Why is the LogicalSource assumed to map to a QueueSource rather than say, a TopicSource?

connectors/activemq/src/main/java/forklift/connectors/ActiveMQConnector.java
```diff
@@ -100,6 +109,31 @@ public ForkliftConsumerI getTopic(String name)
     }
 
     @Override
+    public ForkliftConsumerI getConsumerForSource(SourceI source) throws ConnectorException {
```
AFrieze: I think some javadoc on the interface would help here.  Explain how implementations should throw an exception if the requested source type is not supported by the connector.

connectors/activemq/src/main/java/forklift/connectors/ActiveMQConnector.java
```diff
@@ -100,6 +109,31 @@ public ForkliftConsumerI getTopic(String name)
...
+    private ForkliftConsumerI getGroupedTopic(GroupedTopicSource source) throws ConnectorException {
+        if (!source.groupSpecified())
+            throw new ConnectorException("No consumer group specified");
+
+        // Sanitize the group and topic names so that group and topic names don't accidentally
+        // mess up ActiveMQ's pattern matching
+        final String groupName = source.getGroup().replaceAll("\\.", "_");
+        final String topicName = source.getName().replaceAll("\\.", "_");
+
+        return getConsumerForSource(new QueueSource("Consumer." + groupName + ".VirtualTopic." + topicName));
```
AFrieze: should this be `new GroupedTopicSource`?

Kuroshii: The idea here is to map `GroupedTopicSource`s to a virtual topic in AMQ.
Since `GroupedTopic`s aren't a first class source in AMQ (at least AFAIK), we have to use a queue with the proper naming convention.

Also if I mapped a `GroupedTopicSource` to a `GroupedTopicSource`, wouldn't that cause infinite recursion?

connectors/activemq/src/main/java/forklift/connectors/ActiveMQConnector.java
Kuroshii: I mapped to a `QueueSource` as a light-weight default. I didn't want to use a topic, since you could have the same role running on multiple instances of forklift and you don't really want all of the roles to receive the message.

I don't know why we'd want to (but that doesn't mean it should be precluded) have multiple consumer groups for role input. I could make that work by mapping here to a `TopicSource`, and then mapping that to a `GroupedTopic` when we go to consume from role input. Is that preferable?

connectors/activemq/src/main/java/forklift/connectors/ActiveMQConnector.java
AFrieze: Ok, I think I follow.

core/src/main/java/forklift/connectors/ForkliftSerializer.java
```diff
@@ -0,0 +1,24 @@
...
+     * @return the byte serialization of the given object
+     */
+    public byte[] serializeForSource(SourceI source, Object object);
+
+    /**
+     * Deserialize the given object from bytes.
+     *
+     * @param <T> the type expected from the deserialized object
+     * @param source the source for which the bytes were destined
+     * @param bytes the bytes to deserizlie
```
AFrieze: spelling: deserizlie

connectors/activemq/src/main/java/forklift/connectors/ActiveMQConnector.java
Kuroshii: Done ([link](https://github.com/dcshock/forklift/pull/128/files#diff-6e43cfc614c44ec4138844f4e4e23f01R29))

AFrieze:
I know we don't have this right now, but I'd really like to see a description of how ReplayES works.  It took me a little while to figure it out just looking at the code. Perhaps a combination of class javadoc and/or a README added to the project?

plugins/retry/src/test/java/forklift/retry/RetryLogBuilderTest.java
```diff
@@ -0,0 +1,283 @@
...
+
+import java.lang.annotation.Annotation;
+import java.util.Arrays;
+import java.util.Base64;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Stream;
+
+public class RetryLogBuilderTest {
```
AFrieze: +1 Tests

PR #132, opened 2017-05-24
Wip/forklift kafka incremental controller

AFrieze:
*Overview*
Bridger discovered that calls to subscribe reset the fetch positions of topics.  Additionally, there are some concerns that the topic partitions assigned for topics which do not change from subscription to subscription may be modified if the subscription occurs during some form of rebalance event.  Although it should be possible to keep the existing approach, it was much simpler to simply move to a controller per topic setup.  We may wish to re-visit this in the future should performance become an issue.

### DIFF
--- a/connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
+++ b/connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
@@ -6,10 +6,8 @@ import forklift.controller.KafkaController;
 import forklift.message.MessageStream;
 import forklift.producers.ForkliftProducerI;
 import forklift.producers.KafkaForkliftProducer;
-
 import io.confluent.kafka.serializers.KafkaAvroDeserializerConfig;
 import io.confluent.kafka.serializers.KafkaAvroSerializerConfig;
-
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.producer.KafkaProducer;
@@ -17,6 +15,8 @@ import org.apache.kafka.clients.producer.ProducerConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.TimeUnit;
 
@@ -31,7 +31,7 @@ public class KafkaConnector implements ForkliftConnectorI {
     private final String groupId;
 
     private KafkaProducer<?, ?> kafkaProducer;
-    private KafkaController controller;
+    private Map<String, KafkaController> controllers = new HashMap<>();
 
     /**
      * Constructs a new instance of the KafkaConnector
@@ -63,7 +63,7 @@ public class KafkaConnector implements ForkliftConnectorI {
         return new KafkaProducer(producerProperties);
     }
 
-    private KafkaController createController() {
+    private KafkaController createController(String topicName) {
         Properties props = new Properties();
         props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaHosts);
         props.put(ConsumerConfig.GROUP_ID_CONFIG, groupId);
@@ -76,19 +76,21 @@ public class KafkaConnector implements ForkliftConnectorI {
         props.put(KafkaAvroDeserializerConfig.SPECIFIC_AVRO_READER_CONFIG, false);
 
         final KafkaConsumer<?, ?> kafkaConsumer = new KafkaConsumer(props);
-        return new KafkaController(kafkaConsumer, new MessageStream());
+        return new KafkaController(kafkaConsumer, new MessageStream(), topicName);
     }
 
     @Override
     public synchronized void stop() throws ConnectorException {
-        try {
-            if (controller != null) {
-                controller.stop(2000, TimeUnit.MILLISECONDS);
-                controller = null;
+
+        controllers.values().parallelStream().forEach(controller -> {
+            try {
+                controller.stop(5000, TimeUnit.MILLISECONDS);
+            } catch (InterruptedException e) {
+                log.error("KafkaConnector interrupted while stopping");
             }
-        } catch (InterruptedException e) {
-            log.error("KafkaConnector interrupted while stopping");
-        }
+        });
+        controllers.clear();
+
         if (kafkaProducer != null) {
             kafkaProducer.close();
             kafkaProducer = null;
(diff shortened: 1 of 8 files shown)


connectors/kafka/src/main/java/forklift/connectors/KafkaConnector.java
```diff
@@ -102,8 +104,12 @@ public ForkliftConsumerI getQueue(String name) throws ConnectorException {
 
     @Override
     public synchronized ForkliftConsumerI getTopic(String name) throws ConnectorException {
-        if (controller == null || !controller.isRunning()) {
-            controller = createController();
+        KafkaController controller = controllers.get(name);
+        if(controller != null && controller.isRunning()){
```
Kuroshii: Nit: spacing

connectors/kafka/src/test/java/forklift/integration/ConsumerActionTests.java
```diff
@@ -0,0 +1,40 @@
...
+        ForkliftProducerI producer = forklift.getConnector().getQueueProducer("forklift-string-topic");
+        sentMessageIds.add(producer.send("message1"));
+        final Consumer c1 = new Consumer(StringConsumer.class, forklift);
+        // Shutdown the consumer after all the messages have been processed.
+        c1.setOutOfMessages((listener) -> {
+            listener.shutdown();
+        });
+        // Start the consumer.
+        c1.listen();
+        Thread.sleep(3000);
```
Kuroshii: Do you need this sleep? I'm fairly certain that `Consumer.listen` is blocking.

Kuroshii:
Assuming this works, this should remove the need for your other bugfix PR?

connectors/kafka/src/test/java/forklift/integration/ConsumerActionTests.java
AFrieze: Yes, the sleep is to ensure that the consumer shuts down before the next consumer starts up.

connectors/kafka/src/test/java/forklift/connectors/KafkaControllerTests.java
```diff
@@ -21,14 +22,19 @@
 import org.mockito.Captor;
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
+
 import java.util.Collection;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
```
Kuroshii: Nit: the added imports don't appear to be used

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -145,32 +114,20 @@ public void start() {
      * @throws InterruptedException if the thread is interrupted
      */
     public void stop(long timeout, TimeUnit timeUnit) throws InterruptedException {
-        kafkaConsumer.wakeup();
-        executor.shutdownNow();
-
         running = false;
+        kafkaConsumer.wakeup();
+       // executor.shutdownNow();
+        executor.shutdown();
```
Kuroshii: Do we even need `shutdown` here? The `kafkaConsumer.wakeup()` call should be sufficient to make the message loop thread terminate.

The difference between the two shutdown methods is that `shutdown` prevents incoming tasks from being queued on the executor (which shouldn't matter here since you only call `start` once on a `KafkaController`), while `shutdown` sends a signal `Thread.interrupt` to it's threads to let them know that they _should_ quit soon.

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -184,6 +141,10 @@ private void controlLoop() {
             throw t;
         } finally {
             running = false;
+            synchronized (this) {
+                //wake up any threads waiting on a control loop operation
+                this.notifyAll();
```
Kuroshii: I don't see any `wait`s to be woken up by this notify?

connectors/kafka/src/test/java/forklift/integration/ConsumerActionTests.java
```diff
@@ -0,0 +1,40 @@
...
+     * Test subscribing to a topic, unsubscribing, then resubscribing to the topic.
+     */
+    @Test
+    public void topicTurnoverTest() throws StartupException, ConnectorException, InterruptedException, ProducerException {
+        Forklift forklift = serviceManager.newManagedForkliftInstance();
+        ForkliftProducerI producer = forklift.getConnector().getQueueProducer("forklift-string-topic");
+        sentMessageIds.add(producer.send("message1"));
+        final Consumer c1 = new Consumer(StringConsumer.class, forklift);
+        // Shutdown the consumer after all the messages have been processed.
+        c1.setOutOfMessages((listener) -> {
```
Kuroshii: Hopefully sending just one message doesn't need a timeout for CI... :crossed_fingers:

connectors/kafka/src/test/java/forklift/integration/ConsumerActionTests.java
Kuroshii: I see... the consumer is done processing messages, but you want to make sure that the `close()` is done being called on the previous `KafkaTopicConsumer` before getting a new `KafkaTopicConsumer` read from, because otherwise you could make the new consumer using the old controller (and old `MessageStream`)?

AFrieze: Yes. Now that we can't add and remove from the same controller this test is a little less relevant although I think it's worth leaving it in.

Kuroshii: Thinking some more, I guess I'd blame this being weird on `Consumer.shutdown` being a signal rather than an synchronous action that completes when the consumer is truly "shutdown" or a asynchronous task with a `CompletableFuture`. A better name for what it's doing is `signalShutdown`. :man_shrugging:

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
AFrieze: I think you meant shutdownNow causes an interrupt?  I originally thought that I wouldn't need either of the shutdown calls as well when I did this refactor, but discovered that the awaitTermination requires a shutdown call of some sort to function.   I'll remove the commented shutdownNow

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
AFrieze: Good catch, with the removal of the add/remove topic methods this is no longer needed.

connectors/kafka/src/main/java/forklift/controller/KafkaController.java
Kuroshii: Yeah, whoops.

Hmm, maybe the designers wanted to be able to be sure that when `awaitTermination` is called that no new threads are going to be executed (because that makes the check more difficult than just a bunch of `Thread.join` calls). Seems good then.

Kuroshii:
Wait, what about supporting the new `GroupedTopic` stuff? Make another PR after we merge both PRs?

The difference is that you can only reuse the given `KafkaController` if it has the same consumer group and the same topic as the `GroupedTopic` being requested? In essence, you'd a map from `GroupedTopicSource` to `KafkaController` (probably need to a `hashCode` to be able to do that though).

PR #134, opened 2017-06-07
Switch to ES REST client for Replay Messages

Kuroshii:
#### Changes:
- Switch from ElasticSearch Transport client to a REST client (a directly REST client, as recommended by elasticsearch)

#### Testing:
- Tested running non-erroring messages, without a "replay-version" set in the replay message, delete messages with the same ID in another `forklift-replay*` index and then index a new document in the proper index (ES 2.4 and 5.5).
- Tested ~100 non-erroring messages processing with the latest `replay-version` and observed that they all end up with the final state of `Complete` (ES 2.4 and 5.5)

#### Old Testing:
- Tested that when there is an error, replay messages still end up in an error state (without the version change, out-of-order messages would overwrite final states)
- Tested that the records indexed on the previous day are properly deleted

---

There were some cantankerous connection issues observed when using the ES Transport client when elasticsearch would error and then come back. There have been a few proposed solutions, but since the same problems don't seem to be happening for the REST client, this seems like the most direct solution.

In addition, @arbiter34 pointed out that [support for the transport client is being phased out of ElasticSearch](http://www.elastic.co/blog/state-of-the-official-elasticsearch-java-clients).

Here's one of the alternative approaches I coded up earlier (but hasn't been tested): https://github.com/dcshock/forklift/compare/replay-server-recovery

### DIFF
--- a/plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
+++ b/plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
@@ -17,6 +17,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
+import java.io.Closeable;
 import java.io.IOException;
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
@@ -27,22 +28,35 @@ import java.util.List;
 import java.util.Map;
 import java.util.HashMap;
 
-public class ReplayESWriter extends ReplayStoreThread<ReplayESWriterMsg> {
+public class ReplayESWriter extends ReplayStoreThread<ReplayESWriterMsg> implements Closeable {
     private static final Logger log = LoggerFactory.getLogger(ReplayES.class);
     private static final ObjectMapper mapper = new ObjectMapper();
 
     private final RestClient restClient;
 
+    /**
+     * Creates a log writer that sends replay information to elasticsearch over REST,
+     * using the given hostname and port {@code 9200}.
+     *
+     * @param hostname the name or address of the host to connect to
+     */
     public ReplayESWriter(String hostname) {
-        this(hostname, 9200, "elasticsearch");
+        this(hostname, 9200);
     }
 
-    // clusterName is retained in constructor for compatibility
-    public ReplayESWriter(String hostname, int port, String clusterName) {
+    /**
+     * Creates a log writer that sends replay information to elasticsearch over REST,
+     * using the given hostname and port.
+     *
+     * @param hostname the name or address of the host to connect to
+     * @param port the connection port
+     */
+    public ReplayESWriter(String hostname, int port) {
         this.restClient = RestClient.builder(new HttpHost(hostname, port, "http"))
-            .setRequestConfigCallback(requestConfig -> requestConfig
-                                                           .setConnectTimeout(3_000)
-                                                           .setSocketTimeout(20_000))
+            .setRequestConfigCallback(requestConfig ->
+                requestConfig
+                    .setConnectTimeout(3_000)
+                    .setSocketTimeout(20_000))
             .setDefaultHeaders(new Header[] {
                  new BasicHeader("Accept", "application/json; charset=utf-8"),
                  new BasicHeader("Content-Type", "application/json; charset=utf-8")
@@ -163,7 +177,8 @@ public class ReplayESWriter extends ReplayStoreThread<ReplayESWriterMsg> {
         }
     }
 
-    public void shutdown() {
+    @Override
+    public void close() {
         if (restClient != null) {
             try {
                 restClient.close();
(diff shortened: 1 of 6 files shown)


dcshock:
While I agree that this is the way we have to go... I just want to note that we started by using the REST interface in ES and it turned out to be extremely slow. Might be worth watching as releases are pushed out.

arbiter34:
@dcshock I'm not sure of your previous implementation with Jest, but I believe there are some defaults in it which would cause slowness with high throughput applications.  

Specifically it looks like Jest defaults to using 2 connections max per route by default (PoolingHttpClientConnectionManager default) and 20 max connections total.

plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
```diff
@@ -1,94 +1,167 @@
...
+                 new BasicHeader("Content-Type", "application/json; charset=utf-8")
+             })
+            .setMaxRetryTimeoutMillis(20_000)
+            .build();
     }
 
     @Override
-    protected void poll(ReplayESWriterMsg t) {
-        final String replayVersion = t.getFields().get("forklift-replay-version");
+    protected void poll(ReplayESWriterMsg replayMessage) {
```
seanmrogers: +1 This is cleaner

plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
```diff
@@ -1,94 +1,167 @@
...
 
-            try {
-                // Index the new information.
-                client.prepareIndex(index, "log")
-                    .setVersion(t.getVersion()).setVersionType(VersionType.EXTERNAL_GTE)
-                    .setId(t.getId()).setSource(t.getFields()).execute().actionGet();
-            } catch (VersionConflictEngineException expected) {
-                log.debug("Newer replay message already exists", expected);
+    private void processVersion3Replay(final ReplayESWriterMsg replayMessage) {
+        String indexDate = replayMessage.getFields().get("first-processed-date");
```
seanmrogers: Maybe we should start a class for forklift managed metadata so we can avoid using strings? Probably doesn't matter in this pr but something to think about.

plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
```diff
@@ -1,94 +1,167 @@
...
+import java.util.HashMap;
 
 public class ReplayESWriter extends ReplayStoreThread<ReplayESWriterMsg> {
     private static final Logger log = LoggerFactory.getLogger(ReplayES.class);
+    private static final ObjectMapper mapper = new ObjectMapper();
 
-    private final TransportClient client;
+    private final RestClient restClient;
 
     public ReplayESWriter(String hostname) {
```
AFrieze: javadoc on this constructor might be nice to identify the defaults used.

plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
```diff
@@ -1,94 +1,167 @@
...
     }
 
     public void shutdown() {
-        if (client != null)
-            client.close();
+        if (restClient != null) {
+            try {
+                restClient.close();
+            } catch (IOException e) {
+                log.error("Coudln't shutdown Elastic REST client", e);
```
seanmrogers: nit - `Coudln't`

plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
```diff
@@ -1,94 +1,167 @@
...
+    private void deleteMessageInIndex(final String id, final String index) {
+        final String endpoint = "/" + index + "/log/" + id;
+        try {
+            restClient.performRequest("DELETE", endpoint);
+        } catch (IOException e) {
+            log.error("Error deleting message with id {} for index {}", id, index,  e);
         }
     }
 
     public void shutdown() {
```
AFrieze: Might be worth having this class implement closeable

plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
Kuroshii: +1, that seems like a good idea

Kuroshii:
These changes are incompatible with the previous version of forklift server, since it requires the REST port now instead of the transport port to be passed in. Since that'll be a major version bump either way, I'll just remove the `clusterName` parameter from the ReplayES constructors and forklift server.


@dcshock @arbiter34 As for the inefficiency of using REST, I looked at this set of benchmarks made by elasticsearch (and linked in that article about removing support for the transport client): https://www.elastic.co/blog/benchmarking-rest-client-transport-client. 

What's really conspicuous there is the inefficiency of large **searches** over REST, which matches our old replay processing strategy which would search over `forklift-replay*`. After the update to `replay-version=3`, we are no longer doing any searches in the replay plugin, so I'm optimistic about our performance after this switch.

Kuroshii:
@dcshock I meant to add in a few changes to this before it was merged - I'll make another PR shortly to do that. It's not broken as-is, but since it's already a incompatible change it seems like a fine time to remove some unused bits, and I'm seeing one spot where the documentation is out of date.

PR #137, opened 2017-09-26
fix connector field name

seanmrogers:
### Changes
* A field name was incorrect, causing retries to fail when searching for `connectorName` which should be `destination-connector`

### Testing
* created some retries for forklift to pick up, confirmed old version failed with `Unable to read result` and that the this version doesn't fail. 

### Please Review
@Kuroshii @dcshock @AFrieze 

Not sure how the versioning of this project works, let me know if I need to make changes.

### DIFF
--- a/plugins/retry/build.sbt
+++ b/plugins/retry/build.sbt
@@ -2,7 +2,7 @@ organization := "com.github.dcshock"
 
 name := "forklift-retry"
 
-version := "2.0"
+version := "2.2"
 
 javacOptions ++= Seq("-source", "1.8")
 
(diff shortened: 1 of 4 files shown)


plugins/retry/build.sbt
```diff
@@ -2,7 +2,7 @@ organization := "com.github.dcshock"
 
 name := "forklift-retry"
 
-version := "2.0"
+version := "2.1"
```
Kuroshii: This should probably be `2.2` since forklift already has a version `2.1`. We just haven't been bumping the version if there are no changes.

PR #138, opened 2017-09-28
Optimize out Expensive ES Query in the Replay Plugin

Kuroshii:
#### Changes
- Add a `first-processed-date` property that persists with messages
- Use that `first-processed-date` property to determine the index for a message; no need to do any deleting since elasticsearch's versioning feature provides reliable upserts for us.

For the most part, messages never leave the index they start in, so this helps optimize for that case - after all, dealing with that unconditionally is very excessive. If we really think it's important to move a message to another index when it's reprocessed, I'd recommend moving that kind of logic to forklift-gui.

### Testing

Ran messages through consumers and observed that the `first-processed-date` property was present in the role message. I changed my local date and retried a message that ended in an errored state and observed that it stayed in the same index. 

Performed the same actions on the previous version of the replay plugin and observed that errored messages were written to the index corresponding to the current date.

### Backward Compatibility

_This is based on some observations from testing._

~~Since messages that are written before the change will not be deleted from their original index and then written to a new "permanent" index when retried, error replay logs that were written with the old version of replay will not be automatically fixed when they are retried, regardless of the success of the message.~~

~~Two easy solutions to this:~~
- ~~If the duplicate logs are unacceptable or inconvenient, deal with any errors before making this change to the replay plugin.~~
- ~~Accept that minor inconvenience and manually fix messages after retrying them. If this is the approach, it should be communicated to avoid accidentally duplicating messages.~~

---

Please Review: @dcshock @AFrieze @seanmrogers

### DIFF
--- a/plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
+++ b/plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
@@ -36,27 +36,10 @@ public class ReplayESWriter extends ReplayStoreThread<ReplayESWriterMsg> {
 
     @Override
     protected void poll(ReplayESWriterMsg t) {
-        final String index = "forklift-replay-" + LocalDate.now().format(DateTimeFormatter.BASIC_ISO_DATE);
-
-        // In order to ensure there is only one replay msg for a given id we have to clean the msg id from
-        // any previously created indexes.
-        try {
-            final SearchResponse resp = client.prepareSearch("forklift-replay*").setTypes("log")
-                .setQuery(QueryBuilders.termQuery("_id", t.getId()))
-                .setFrom(0).setSize(50).setExplain(false)
-                .execute()
-                .actionGet();
-
-            if (resp != null && resp.getHits() != null && resp.getHits().getHits() != null) {
-                for (SearchHit hit : resp.getHits().getHits()) {
-                    if (!hit.getIndex().equals(index))
-                        client.prepareDelete(hit.getIndex(), "log", t.getId()).execute().actionGet();
-                }
-            }
-        } catch (Exception e) {
-            log.error("", e);
-            log.error("Unable to search for old replay logs {}", t.getId());
-        }
+        String indexDate = t.getFields().get("first-processed-date");
+        if (indexDate == null)
+            indexDate = LocalDate.now().format(DateTimeFormatter.BASIC_ISO_DATE);
+        final String index = "forklift-replay-" + indexDate;
 
         try {
             // Index the new information.
--- a/plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
+++ b/plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
@@ -36,18 +36,54 @@ public class ReplayESWriter extends ReplayStoreThread<ReplayESWriterMsg> {
 
     @Override
     protected void poll(ReplayESWriterMsg t) {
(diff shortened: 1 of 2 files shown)


AFrieze:
Could the Backward Compatibility be improved through some use of the stepCount and having the deletion logic run conditionally?

Kuroshii:
We could improve backwards compatibility by adding the replay-version field to message properties and then just performing "the old thing" if we aren't on the new version.

On the other hand, that complicates the update and I think the consequences aren't dire.

Kuroshii:
@AFrieze Updated for backward compatibility - testing it now.

plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
```diff
@@ -36,35 +36,54 @@ public ReplayESWriter(String hostname, int port, String clusterName) {
...
-                .actionGet();
+            try {
+                // Index the new information.
+                client.prepareIndex(index, "log")
+                    .setVersion(t.getVersion()).setVersionType(VersionType.EXTERNAL_GTE)
+                    .setId(t.getId()).setSource(t.getFields()).execute().actionGet();
+            } catch (VersionConflictEngineException expected) {
+                log.debug("Newer replay message already exists", expected);
+            }
+        } else if ("unknown".equals(replayVersion)) { // manage old versions, for backward compatibility
```
seanmrogers: Any chance that this could be `null`? Should probably handle it with an else case JIC?

Kuroshii:
There's some non-backward compatible code that's should be changed in forklift-gui for this to work. The order of the upgrades doesn't matter, but forklift GUI won't correctly retry anything other than version `2` at the moment - PR here: https://github.com/dcshock/forklift-gui/pull/20

Tested reading "newly retried" and "previously retried" messages on both versions of the replay plugin and verified that version `2` messages change index and version `3` messages don't change index.

plugins/replay/src/main/java/forklift/replay/ReplayESWriter.java
```diff
@@ -36,35 +36,54 @@ public ReplayESWriter(String hostname, int port, String clusterName) {
...
-            final SearchResponse resp = client.prepareSearch("forklift-replay*").setTypes("log")
-                .setQuery(QueryBuilders.termQuery("_id", t.getId()))
-                .setFrom(0).setSize(50).setExplain(false)
-                .execute()
-                .actionGet();
+            try {
+                // Index the new information.
+                client.prepareIndex(index, "log")
+                    .setVersion(t.getVersion()).setVersionType(VersionType.EXTERNAL_GTE)
+                    .setId(t.getId()).setSource(t.getFields()).execute().actionGet();
```
AFrieze: Do you need the Get here?

plugins/replay/src/main/java/forklift/replay/ReplayLogBuilder.java
```diff
@@ -65,8 +66,13 @@ public ReplayLogBuilder(ForkliftMessage msg, Consumer consumer, List<String> err
         // carry a description of the original source across restarts
         props.putIfAbsent("source-description", sourceDescription);
 
+
+        if (!props.containsKey("forklift-replay-step-count")) { // new messages are version 3
+            props.putIfAbsent("forklift-replay-version", "3");
+        }
```
AFrieze: Do you need to put forklift-replay-version of 2 here if you there is a step count present but no version?

PR #139, opened 2017-10-16
Kafka Connector: Consumer Topic + Grouped Topic Support

Kuroshii:
#### Changes:
- Index `KafkaController` instances based on both topic name and the chosen groupId for a topic this way, each different `GroupedTopicSource`s gets it's own underlying `KafkaConsumer`
- If class-annotated, a `TopicSource` is identified by both the default groupId and it's class's name
- Add tests for the way sources are mapped

I ran all the tests locally and they all _eventually_ passed after a rerun and 2GB of memory. Having another look at our integration tests is a good idea.

---
Please Review: @seanmrogers @dkessler-sofi @AFrieze


dcshock:
@Kuroshii can you add a release note to the README and update the version for release? Thanks

dcshock:
@Kuroshii is this something that we'd like to get merged in? I'm not familiar enough with the kafka connector to make the assessment.

PR #140, opened 2017-11-20
Feature/use reader schema for specific records

AFrieze:
The KafkaAvroDeserializer uses the writer schema retrieved from the incoming message to deserialize messages into GenericRecords.  This leads to default values in a newer Reader schema being ignored.  This PR adds a new deserializer that is configured with the appropriate reader schema based off a source.  It is worth noting that this PR removes the `getQueue(String name)` and `getTopic(String name)`  from the ForkliftConnector interface.


core/src/main/java/forklift/connectors/ForkliftConnectorI.java
```diff
@@ -33,32 +33,10 @@
...
-    /**
-     * Retrieves the {@link ForkliftConsumerI consumer} instance that reads from
-     * the queue with the given name.
-     *
-     * @param name the name of the topic to read from
-     * @return a consumer that reads from the given topic
-     * @throws ConnectorException if an error occurred interacting with the connector
-     * @throws RuntimeException if reading from a queue is not supported
-     */
-    ForkliftConsumerI getQueue(String name) throws ConnectorException;
```
dcshock: Might be good to stub these in to call getConsumerForSource to avoid backward compatibility issues.

AFrieze: How much of a concern is this given that these methods are mostly internal?  My rationale for removing these is that more and more functionality is moving into the "Source" concept and if keep these methods around we could have diverging logic in places.

dcshock:
It looks like we forced this into develop. @Kuroshii @AFrieze can this PR be closed?

PR #143, opened 2017-12-19
Add missing heartbeat to kafka controller

AFrieze:


connectors/kafka/src/main/java/forklift/controller/KafkaController.java
```diff
@@ -186,8 +186,9 @@ private void controlLoop() {
             synchronized (this) {
                 //wait for flowControl to notify us, resume after a short pause to allow for heartbeats
                 this.wait(100);
+                //Heartbeat is sent on poll. This call should not return records.
+                return kafkaConsumer.poll(1);
```
Kuroshii: For this, `poll(0)` should be better (assuming this is supposed to be avoiding a busy loop). From the javadoc for `KafkaConsumer#poll`:
```
     * @param timeout The time, in milliseconds, spent waiting in poll if data is not available in the buffer.
     *            If 0, returns immediately with any records that are available currently in the buffer, else returns empty.
     *            Must not be negative.
```

The heartbeat that is triggered by polling the `ConsumerCoordinator` is completely asynchronous (it is sent from the coordinator's heartbeat thread), so it is realistic that `poll(0)` might do a few loops through `pollOnce`.

PR #145, opened 2018-03-07
Fix Kafka Processing Errors

Kuroshii:
### Summary
- Handle rejected executions by blocking to enqueue the rejected message.
- Invalidate records based on a new feature - generation number
- Shutdown consumers faster by emptying the work queue first
- Add test case for duplicate processing records

### Overview
We are seeing two different kinds of undesired processing: missed messages, and duplicate messages. Below I describe my current reasoning for what is causing these issues and how it can be fixed.

### Missed Messages
The missed messages seem to be caused by out-of-order processing created by the `CallerRunsPolicy` used for handling messages that can't be queued on the `Consumer` side, combined with a large consumer queue that can't be fully processed in a normal shutdown sequence.

My solution for this is to have rejected executions wait to be queued on the consumer's thread. (An alternative policy would be to have the caller process a message at the head of the work queue, but I prefer to keep the processing in the consumer's thread pool.)

### Duplicate Messages
The duplicate messages are most likely caused by incorrect behavior around rebalancing, where stale messages are kept after a rebalance. (This has been validated by the new test case.)

There are two general approaches here:
- Only let valid records exit `MessageStream` (requiring truly ordered consumption on a partition) and removing cached records as soon as partitions are revoked.
- Filter out invalid messages when the consumer first processes them, having some way of indicating the oldness of a message

Although the first option is probably better practice, the second option is quicker to write and lets us avoid risky architectural changes.

#### Solution
This PR introduces a "generation" managed by `KafkaController` for each topic-partition, and updated as partitions are assigned and revoked. When a message is created, this generation number is captured, and if it doesn't match the latest generation number for the message's partition that message is skipped by `KafkaMessage#acknowledge`.

---

Please Review: @AFrieze @dcshock @seanmrogers

### DIFF
--- a/core/src/main/java/forklift/concurrent/BlockingRejectedExecutionHandler.java
+++ b/core/src/main/java/forklift/concurrent/BlockingRejectedExecutionHandler.java
@@ -1,0 +1,20 @@
+package forklift.concurrent;
+
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.RejectedExecutionHandler;
+
+/**
+ * A rejected execution handler for a {@link ThreadPoolExecutor} that handles rejected
+ * tasks from a running thread pool by blocking.
+ */
+public class BlockingRejectedExecutionHandler implements RejectedExecutionHandler {
+    @Override
+    public void rejectedExecution(Runnable runnable, ThreadPoolExecutor executor) {
+        while (!executor.isShutdown()) {
+            try {
+                executor.getQueue().put(runnable);
+                return;
+            } catch (InterruptedException ignored) {}
+        }
+    }
+}
--- a/core/src/main/java/forklift/concurrent/BlockingRejectedExecutionHandler.java
+++ b/core/src/main/java/forklift/concurrent/BlockingRejectedExecutionHandler.java
@@ -8,13 +8,20 @@ import java.util.concurrent.RejectedExecutionHandler;
  * tasks from a running thread pool by blocking.
  */
 public class BlockingRejectedExecutionHandler implements RejectedExecutionHandler {
+    private static final Runnable NO_OP = () -> {};
+
     @Override
     public void rejectedExecution(Runnable runnable, ThreadPoolExecutor executor) {
-        while (!executor.isShutdown()) {
-            try {
-                executor.getQueue().put(runnable);
-                return;
-            } catch (InterruptedException ignored) {}
-        }
+        if (executor.isShutdown()) { return; }
+
+        try {
+            // wait for space in the queue
+            executor.getQueue().put(NO_OP);
+            executor.remove(NO_OP);
+
+            if (!executor.isShutdown()) {
+                executor.execute(runnable);
+            }
+        } catch (InterruptedException ignored) {}
     }
 }
--- a/connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
+++ b/connectors/kafka/src/main/java/forklift/controller/AcknowledgedRecordHandler.java
@@ -12,6 +12,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Supplier;
 
 /**
  * Maintains a batch of acknowledged records and provides management for adding and removing {@link org.apache.kafka.common.TopicPartition partitions}.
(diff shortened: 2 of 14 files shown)


Kuroshii:
Found a race condition happening in Travis tests, where a message would be successfully accepted for it's generation and then have it's offsets merged after a rebalance, causing duplicate processing.

I refactored the code a bit to have the generation checks happen atomicly with the merging of offsets, which should prevent this from being an issue.

seanmrogers:
Perhaps we should have some more logging in place? In case we have to debug?

Kuroshii:
@seanmrogers There is actually a fair amount of logging already there, it's just only visible if you have logging configured to display forklift's debug messages.

Is it sufficient to have debug logs post message acknowledgment and include generation numbers on rebalancing? Or are you thinking of something else?

Patch of what I was thinking:
```java
diff --git a/connectors/kafka/src/main/java/forklift/controller/KafkaController.java b/connectors/kafka/src/main/java/forklift/controller/KafkaController.java
index f4a8421..2226b4e 100644
--- a/connectors/kafka/src/main/java/forklift/controller/KafkaController.java
+++ b/connectors/kafka/src/main/java/forklift/controller/KafkaController.java
@@ -237,7 +237,7 @@ public class KafkaController {
     private class RebalanceListener implements ConsumerRebalanceListener {
         @Override
         public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
-            log.debug("controlLoop partitions revoked");
+            log.debug("controlLoop partitions revoked with generations {}", generations);
             try {
                 for (TopicPartition partition : partitions) {
                     generations.get(partition).unassign();
@@ -264,6 +264,8 @@ public class KafkaController {
                 generation.assignNextGeneration();
             }

+            log.debug("controlLoop partitions assigned with generations {}", generations);
+
             acknowledgmentHandler.addPartitions(partitions);
             for (TopicPartition partition : partitions) {
                 flowControl.merge(partition, new AtomicInteger(),
diff --git a/connectors/kafka/src/main/java/forklift/message/KafkaMessage.java b/connectors/kafka/src/main/java/forklift/message/KafkaMessage.java
index 93376ee..40ac2da 100644
--- a/connectors/kafka/src/main/java/forklift/message/KafkaMessage.java
+++ b/connectors/kafka/src/main/java/forklift/message/KafkaMessage.java
@@ -8,8 +8,12 @@ import forklift.producers.KafkaForkliftProducer;

 import org.apache.avro.generic.GenericRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;

 public class KafkaMessage extends ForkliftMessage {
+    private static final Logger log = LoggerFactory.getLogger(KafkaMessage.class);
+
     private final KafkaController controller;
     private final ConsumerRecord<?, ?> consumerRecord;
     private final long generationNumber;
@@ -29,7 +33,10 @@ public class KafkaMessage extends ForkliftMessage {
     @Override
     public boolean acknowledge() throws ConnectorException {
         try {
-            return controller.acknowledge(consumerRecord, generationNumber);
+            final boolean acknowledged = controller.acknowledge(consumerRecord, generationNumber);
+            log.debug("Acknoledgment for topic: {} partition: {} offset: {} generation: {} successful: {}",
+                      consumerRecord.topic(), consumerRecord.partition(), consumerRecord.offset(), generationNumber, acknowledged);
+            return acknowledged;
         } catch (InterruptedException e) {
             throw new ConnectorException("Error acknowledging message");
         }
```

core/src/main/java/forklift/concurrent/BlockingRejectedExecutionHandler.java
```diff
@@ -0,0 +1,27 @@
...
+    private static final Runnable NO_OP = () -> {};
+
+    @Override
+    public void rejectedExecution(Runnable runnable, ThreadPoolExecutor executor) {
+        if (executor.isShutdown()) { return; }
+
+        try {
+            // wait for space in the queue
+            executor.getQueue().put(NO_OP);
+            executor.remove(NO_OP);
```
AFrieze: I'm not too familiar with RejectedExecutionHandlers but, why not add the runnable directly to the queue?  Is it to avoid the case where you add to the queue while the executor is shutting down?  Also, would it be quicker to just allow the NO_OP to run rather than remove it?

Kuroshii: >  Is it to avoid the case where you add to the queue while the executor is shutting down?

Yup. 

Since there's no way (to my knowledge) a generic queue itself can stop from receiving elements, using `put` you could get out-of-order execution doing something normally reasonable like:
```java
pool.shutdown(); // stop accepting tasks, set the pool to be on the path to being terminated
pool.getQueue().clear(); // remove any old tasks
/* probable context switch (since blocking is implemented with wait-notify) 
   -- waiting element gets queued */
pool.awaitTermination(...); // wait

pool.shutdownNow(); // interrupt any operations, there's no more time!
```

Basically, adding a harmless element is an easy way to wait for a `BlockingQueue` to have space. The other alternative is to do a poll loop on `BlockingQueue#remainingCapacity` without a wait-notify.

Realistically, this particular implementation isn't a huge concern since the rejected execution handler runs on the same thread, so a shutdown can't happen in the middle of the rejected execution handler. If you don't like it, I can just do `executor.getQueue().put(action)` and then throw it in a subclass of `Consumer` to hide it from misuse (after all, it could probably still be misused as-is by having multiple threads feed the same thread pool).

> Also, would it be quicker to just allow the NO_OP to run rather than remove it?

Huh? It should be at the back end of an almost full queue, so that seems like that would be equivalent to waiting for all of the elements in the initial queue (before waiting) to complete?

seanmrogers:
@Kuroshii Yes, I think logging post message acknowledgment and include generation numbers on rebalancing would be helpful debug logs.

seanmrogers:
LGTM

PR #148, opened 2018-05-10
Modify consumer method annotations to be detected on superclasses

yashton:
For consumers with a high degree of shared boilerplate on consumer methods, creating a base class to abstract the boilerplate fails because the annotations only are detected on direct members of the consumer. Allowing forklift to detect annotations on super classes and interfaces solves this issue.


dcshock:
@applitect I completely missed this as I haven't been the maintainer in quite some time. We should definitely get this in place, but it'll require a quick update to freshen it up with the latest dev version. Is this something you can take on?

applitect:
This branch is not valid. Methods and fields are not supported by @Inherited. That is only supported for classes. Thus the unit tests in OnDecorator fail and the methods can not be found for things like OnMessage, etc.

PR #150, opened 2018-05-23
Perform Acknowledgment After Processing When Working with AMQ

Kuroshii:
#### Changes:
- Split `ForkliftMessage` acknowledgment into two pieces, one which runs beforehand and can signal messages to be skipped and the other which runs after and is expected to always succeed.

#### Release Note:
- Made ActiveMQ consumers acknowledge messages after they are processed so as to provide at-least-once processing and avoid dropping messages.

#### Testing:
- Tested replay and retry for both AMQ and forklift using a short timeout (10 seconds).
    - Noted that messages themselves go over the same queues (the original queue initilaly, and then the role queue for any subsequent messges) and get consumed properly (testing the changes to acknowledging message).
    - Noted that the proper entries appear in `forklift-replay` and `forklift-retry` in elasticsearch. That is, the replay index had a message with the message's ID and a status of `RETRYING` while the message was being retried, and a status of `ERROR` after that. In the retry index there was a log in the retry index for the corresponding message ID only when the message was being retried.
    - For the kafka-connector, with debug logging enabled, I noted that acknowledgment messges re indeed printed before the message is processed.
    
- Tested that acknowledgments after processing protect from spontaneous crashes in ActiveMQ. 
   - Verified that the if I force quit the application in the consumer's long processing time, the message doesn't get acked (still shows up in the admin UI), and then gets reprocessed when the application starts up again.
    - Verified that the application does get acked and not reprocessed in normal (non-crashing) processing.

- Used this consumer code on forklift-server (ActiveMQ) for ack testing:
```java
@MultiThreaded(3)
@Queue("test")
@Replay
@RoleInput
@Retry(maxRetries=5, timeout=10)
public class TestConsumer {
    private static final Logger log = LoggerFactory.getLogger(TestConsumer.class);

    @Message public String message;

    @OnMessage
    public void onMessage() {
        log.info("Starting test processing: {}", message);

        try {
            Thread.sleep(10000);
        } catch (final InterruptedException e) {}

        log.info("Finished test processing: {}", message);
    }
}
```

After considerable discussion with @AFrieze and @nateww, I think it makes sense to say a few things to justify these changes:

- The typical and original usage of AMQ with forklift was to perform a piece of work reliably for a business process, where missing a message means that a process is unfulfilled.
- Crashes in most applications are inevitable (although their rarity can be debated).
- The change from acknowledging post-processing to pre-processing was made haphazardly, without considering missed messages.
- The kafka connector doesn't fit cleanly into this with the current threading model, and has been documented to be have an at-least-once processing guarantee.

Thus my conclusion is to let the Kafka connector continue to perform at-most-once processing, but change how AMQ is used to work with the original expectation of at-least-once processing. 

---

Please Review: @dcshock @AFrieze @nateww

### DIFF
--- a/core/src/main/java/forklift/consumer/MessageRunnable.java
+++ b/core/src/main/java/forklift/consumer/MessageRunnable.java
@@ -61,15 +61,15 @@ public class MessageRunnable implements Runnable {
     public void run() {
         RunAsClassLoader.run(classLoader, () -> {
             // Always ack message to prevent deadlocks
-            boolean acknowledged = false;
+            boolean continueProcessing;
             try {
-                acknowledged = msg.acknowledge();
+                continueProcessing = msg.beforeProcessing();
             } catch (ConnectorException e) {
-                log.error("Error while acking message.", e);
-                acknowledged = false;
+                continueProcessing = false;
+                log.error("Error before processing message", e);
             }
 
-            if (!acknowledged) {
+            if (!continueProcessing) {
                 close();
                 return;
             }
@@ -155,6 +155,17 @@ public class MessageRunnable implements Runnable {
                     this.lifeCycle.call(ProcessStep.Complete, this);
                 }
             }
+
+            // acknowledge only after all the work has been said and done
+            try {
+                final boolean acknowledged = msg.acknowledge();
+                if (!acknowledged) {
+                    log.error("Failed to acknowledge message after processing");
+                }
+            } catch (Exception e) {
+                log.error("Error while acking message", e);
+            }
+
             // Always log all non-null errors
             if (this.warnOnly)
                 getErrors().stream().filter(e -> e != null).forEach(e -> log.warn(e));
--- a/connectors/kafka/src/main/java/forklift/message/KafkaMessage.java
+++ b/connectors/kafka/src/main/java/forklift/message/KafkaMessage.java
@@ -31,7 +31,7 @@ public class KafkaMessage extends ForkliftMessage {
     }
 
     @Override
-    public boolean acknowledge() throws ConnectorException {
+    public boolean beforeProcessing() throws ConnectorException {
         try {
             final boolean acknowledged = controller.acknowledge(consumerRecord, generationNumber);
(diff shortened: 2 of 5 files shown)


core/src/main/java/forklift/consumer/MessageRunnable.java
```diff
@@ -155,6 +155,17 @@ public void run() {
                     this.lifeCycle.call(ProcessStep.Complete, this);
```
AFrieze: Do you want the complete step to run before the final acknowledgment of message receipt?

BirdHowl: Good catch - probably not. 

I don't think we'd want plugins to do things as if everything was successful and then have the acknowledgment fail (in case the plugin isn't idempotent wrt to the lifecycle of a message).

That precludes hooks from skipping the acknowledgment if something fails, but that use case doesn't really exist yet.

BirdHowl: Thought about this some more, and changed my mind.

It's non-obvious (to me) what should happen for `Error` and `Invalid` when acknowledgment fails.

It could maybe work like this:
```
Validating -> Invalid => DoAck()
Validating -> Processing -> Error => DoAck()
Validating -> Processing => DoAck() -> Complete
```

That's probably the "most useful" placement of acknowledgments, but it's also more complicated to understand.

I think it makes sense to say that since messages can be reprocessed, it's also possible for any of the lifecycle events to occur again for a message when that happens, and so plugins should be idempotent if duplicate processing is a concern.

Especially since failed acknowledgment should be very rare, and I know of don't know of need for a plugin to operate only in conditions where the acknowledgment has succeeded, I favor the simpler approach we have currently.

nateww: I think that the Ack should happen after everything has completed, and the message is fully processed, which is the most common means for an ACK in all protocols I've been involved with.  Plus, the original code prior to Todd's changes had it there as well, so I believe that was Matt's intent.

PR #151, opened 2018-06-01
Try to make publishing not fail if there is a bad artifact staged

Kuroshii:


Kuroshii:
Tested on master and was able to successfully release version 3.5. Merging in.

PR #154, opened 2020-07-16
Convert to gradle

applitect:
- Rewrite the build framework to go from sbt to gradle. 
- Update to release with Java 11.
- Minor javadoc fixes to allow it to compile with Java 11.

Has been tested and have been able to deploy a snapshot version of 3.8 to sonatype oss. Deploys in the same structure as the sbt deploy.


connectors/activemq/build.gradle
```diff
@@ -0,0 +1,9 @@
+dependencies {
+    api project(':forklift-core'),
+            'org.apache.activemq:activemq-client:5/14.0',
```
Kuroshii, line 3: Should this be `5.14.0`?

applitect: Probably. I bet in came in transitively from plug-in. Going to look to see if I can remove it.

applitect: I bet bringing in the broker brought in the client correctly. But fixing.

.travis.yml
```diff
@@ -1,8 +1,6 @@
-language: scala
-scala:
-- 2.11.4
+language: java
 jdk:
-- oraclejdk8
```
dcshock, line 5: Let's be sure to note this in the release notes. "No longer support jdk8"

connectors/activemq/build.gradle
```diff
@@ -0,0 +1,9 @@
+dependencies {
+    api project(':forklift-core'),
+            'org.apache.activemq:activemq-client:5.14.0',
+            'org.apache.activemq:activemq-broker:5.14.0',
```
dcshock, line 4: Do we want to push the broker code to everyone that uses the connector library?

server/build.gradle
```diff
@@ -0,0 +1,34 @@
+apply plugin: 'java'
+apply plugin: 'application'
```
dcshock, line 2: Does this build the same binary as the sbt version?

.travis.yml
applitect: Already got another branch going for updating docs for the next release.

connectors/activemq/build.gradle
applitect: Good question. Let me look at that.

server/build.gradle
applitect: Yes, it builds the same binary and creates a shell script. Not identical shell script but it works the same.

PR #159, opened 2020-08-05
Upgrade Kafka Versions

Kuroshii:
**Changes:**

- Upgrade kafka and confluent versions to the latest
- Use the latest versions of kafka, zookeeper, schema-registry in tests
- Remove deprecated kafka code, upgrade `TestConsumer` to work with Kafka 2.5


build.gradle
```diff
@@ -16,6 +16,9 @@ allprojects {
         maven {
             url 'https://packages.confluent.io/maven'
         }
+        maven {
```
applitect, line 19: Why the separate maven block?

PR #163, opened 2020-09-03
Allow named injections

applitect:


core/src/main/java/forklift/consumer/Consumer.java
```diff
@@ -401,7 +402,9 @@ private final void configureConstructorInjection() {
                 fields.get(clazz).forEach(field -> {
                     log.trace("Inject target> Field: ({})  Decorator: ({})", field, decorator);
                     try {
-                        Object value = getInjectableValue(field.getAnnotation(decorator), field.getName(), clazz, msg);
+                        String fieldName = Optional.ofNullable(field.getAnnotation(Named.class))
+                                .map(a -> a.value()).orElse(field.getName());
```
adispennette, line 406: do you want orElse or orElseGet?  orElse is always executed where as orElseGet is not

adispennette (approved):

core/src/test/java/forklift/consumer/injection/ConsumerTest.java
```diff
@@ -224,6 +229,49 @@ public void testHeadersAndProperties() {
         assertEquals(ec.producer, "testing");
     }
 
+    // For the next two tests, we are testing named dependency injection.
```
AFrieze, line 232: Minor Opinion:  This comment seems a little brittle in that code methods could move

core/src/test/java/forklift/consumer/injection/ServiceNamedBeanResolver.java
```diff
@@ -0,0 +1,23 @@
...
+
+import forklift.decorators.BeanResolver;
+import forklift.decorators.Service;
+
+@Service
+public class ServiceNamedBeanResolver {
+    static Map<String, Person> map = new HashMap<>();
+
+    @BeanResolver
+    public Object resolve(Class<?> requestedClass, String name){
```
AFrieze, line 14: Does scope matter at all?  It looks like if we have a named bean then we are applying `Singleton` scope and if we have a non-named bean we are creating new instances on every request

AFrieze (approved):

core/src/test/java/forklift/consumer/injection/ServiceNamedBeanResolver.java
applitect: This is fine, since it it only being used in the tests to prove that the named annotation resolver can be used successfully. This is not production level code.

